PR #42, opened 2019-03-25
Update App test to test for finding elements

jacobvenable:
This PR is not associated with an issue.

## What does this PR do?

Adds an initial test for the `<App/>` component using [Enzyme's shallow rendering](https://airbnb.io/enzyme/docs/api/shallow.html) to find a certain number of components.

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import App from './App';
 import { Route } from 'react-router-dom';
+import App from './App';
 import Header from './components/Header';
 import Home from './components/Home';
 import Shelters from './components/Shelters';
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -1,7 +1,21 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import App from './App';
+import { Route } from 'react-router-dom';
+import Header from './components/Header';
+import Home from './components/Home';
+import Shelters from './components/Shelters';
+import Hygiene from './components/Hygiene';
+import Hotlines from './components/Hotlines';
+import Food from './components/Food';
+import Transit from './components/Transit';
 
-it('renders without crashing', () => {
-  shallow(<App />);
+describe('<App />', () => {
+  const wrapper = shallow(<App />);
+  const routes = [Home, Shelters, Hygiene, Hotlines, Food, Transit];
+  it('renders properly', () => {
+    expect(wrapper.find('div').length).toBe(1);
+    expect(wrapper.find(Header).length).toBe(1);
+    expect(wrapper.find(Route).length).toBe(routes.length);
+  });
 });
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -1,21 +1,46 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import { Route } from 'react-router-dom';
 import App from './App';
-import Header from './components/Header';
-import Home from './components/Home';
-import Shelters from './components/Shelters';
-import Hygiene from './components/Hygiene';
-import Hotlines from './components/Hotlines';
-import Food from './components/Food';
-import Transit from './components/Transit';
+jest.mock('./components/Header', () => 'Header');
+jest.mock('react-router-dom/Route', () => 'Route');
 
 describe('<App />', () => {
   const wrapper = shallow(<App />);
-  const routes = [Home, Shelters, Hygiene, Hotlines, Food, Transit];
+  const routes = [
+    {
+      path: '/',
+      component: 'Home'
+    },
+    {
+      path: '/shelters',
+      component: 'Shelters'
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/App.test.js
```diff
@@ -1,7 +1,21 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import App from './App';
+import { Route } from 'react-router-dom';
```
kyleruschel: As a best practice, you should always have absolute imports before relative. Easy fix, feel free to reach out for more clarity 👍

jacobvenable: Updated in latest commit.

jacobvenable:
@kyleruschel sorry, `<App/>` pre-existed the test in this case but I made an insignificant update to it so you can see the file in the latest commit. New components I've built will be truly be done via TDD.

src/App.test.js
```diff
@@ -1,7 +1,21 @@
...
+import Food from './components/Food';
+import Transit from './components/Transit';
 
-it('renders without crashing', () => {
-  shallow(<App />);
+describe('<App />', () => {
+  const wrapper = shallow(<App />);
+  const routes = [Home, Shelters, Hygiene, Hotlines, Food, Transit];
+  it('renders properly', () => {
+    expect(wrapper.find('div').length).toBe(1);
```
kyleruschel: The check of the parent `div` here can be omitted, since by default you must have a parent `div` when returning more than one element. For this practice, I see no issue to check because you're verifying that it does actually exist but your component would never render without it so it's quite impossible to miss. 👍

src/App.test.js
```diff
@@ -1,7 +1,21 @@
 import React from 'react';
 import { shallow } from 'enzyme';
+import { Route } from 'react-router-dom';
 import App from './App';
+import Header from './components/Header';
+import Home from './components/Home';
```
kyleruschel: So, here is something that we've recently started to implement that will be beneficial for not just not but future as well. We can mock these imports by using ```jest.mock('./components/Header', () => 'Header');``` in lieu of your imports and it will achieve the same results 👍 . Reach out to `Gordon` for further clarity...

src/App.test.js
jacobvenable: Although I haven't used them much, couldn't I also use a [fragment](https://reactjs.org/docs/fragments.html) instead?

src/App.test.js
jacobvenable: @gordonhgraham

src/App.test.js
kyleruschel: You could, yes but I don't believe `<Fragment>` is 100% fully supported so I tend to still use the old trusty `<div>'. I could be incorrect on this, but I want to say I've read where `<>` is also viable, but check and get back to me on this..

gordonhgraham: i agree @kyleruschel, @jacobvenable by checkign that i can see you're writing tests first, once youv'e verified somthing that would require that first test to be true you can remove it in a refactor....`Red`, `Green`, `Refactor` 👍  👏

src/App.test.js
gordonhgraham: good call out @kyleruschel, check out the `jest` documentation on mocks for the syntax.

The reason we do that is these unit tests don't care about the specific `header` component. they just care that it exists.

src/App.test.js
```diff
@@ -1,7 +1,45 @@
...
+      path: '/food',
+      component: 'Food'
+    },
+    {
+      path: '/transit',
+      component: 'Transit'
+    }
+  ];
+  it('renders properly', () => {
+    expect(wrapper.find('Header').length).toBe(1);
```
gordonhgraham: i'd pull these two expectations out into two separate `it` blocks...

```
it('renders the header component', () => {
  ...logic for checking header...
})
```
```
it('matches each route w/ appropriate component', () => {
  ...logic for routes...
})
```

src/App.test.js
```diff
@@ -1,7 +1,45 @@
...
+      path: '/hotlines',
+      component: 'Hotlines'
+    },
+    {
+      path: '/food',
+      component: 'Food'
+    },
+    {
+      path: '/transit',
+      component: 'Transit'
```
gordonhgraham: good call on keeping the components as strings since we don't care about the exact components.

src/App.test.js
```diff
@@ -1,7 +1,45 @@
...
+    }
+  ];
+  it('renders properly', () => {
```
kyleruschel: I would update this block to be more descriptive of what you're trying to test. Remember tests should almost read like a book, dictating what you're trying to accomplish.

src/App.test.js
```diff
@@ -1,7 +1,45 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import App from './App';
+jest.mock('./components/Header', () => 'Header');
+jest.mock('react-router-dom/Route', () => 'Route');
```
gordonhgraham: 👌 👍

PR #43, opened 2019-03-25
Add/landing page/search initial

jacobvenable:
This PR is not tied to a specific issue.

## What does this PR do?

Adds the initial version of `<Search/>` and adds it to the home page.

- `<Search/>` was created using some simple TDD to first generate.
- `<Home/>` test was created after the component was already created but had it's returned JSX completely rewritten via TDD.

### Questions/Topics

Is it necessary to test that the same number of a layout component are necessary? In the `Home.test.js`, I test that the proper number [Grid components](https://material-ui.com/layout/grid/#grid) is correct, but that can easily change in the future.

Maybe a snapshot should be used instead?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6703,7 +6703,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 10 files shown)


jacobvenable:
@kyleruschel do you think it's necessary to test for the existence of `<Grid/>` components?

PR #44, opened 2019-03-26
Add/landing page/buttons

jacobvenable:
This PR is not associated with a specific issue.

## What does this PR do?

Add both major and minor buttons to the home page. Buttons marked as major will eventually be more visually significant and then be followed by less-significant minor buttons.

### DIFF
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -1,11 +1,115 @@
 import React from 'react';
-import { Grid } from '@material-ui/core';
+import { Grid, IconButton } from '@material-ui/core';
+import CallIcon from '@material-ui/icons/Call';
+import RestaurantIcon from '@material-ui/icons/Restaurant';
+import HomeIcon from '@material-ui/icons/Home';
+import HotTubIcon from '@material-ui/icons/HotTub';
+import BusIcon from '@material-ui/icons/DirectionsBus';
+import GroceryStoreIcon from '@material-ui/icons/LocalGroceryStore';
+import HealingIcon from '@material-ui/icons/Healing';
+import LibraryBooksIcon from '@material-ui/icons/LibraryBooks';
+import PetsIcon from '@material-ui/icons/Pets';
+import WifiIcon from '@material-ui/icons/Wifi';
+import WorkIcon from '@material-ui/icons/Work';
+import InfoIcon from '@material-ui/icons/Info';
 import Search from './Search';
 
-const Home = () => (
+export const HomeButtonsMajor = [
+  {
+    text: 'Hotlines',
+    icon: CallIcon
+  },
+  {
+    text: 'Food',
+    icon: RestaurantIcon
+  },
+  {
+    text: 'Shelter',
+    icon: HomeIcon
+  },
+  {
+    text: 'Hygiene',
+    icon: HotTubIcon
+  },
+  {
+    text: 'Transit',
+    icon: BusIcon
+  },
+  {
+    text: 'Resources',
+    icon: GroceryStoreIcon
+  }
+];
+
+export const HomeButtonsMinor = [
+  {
+    text: 'Health',
+    icon: HealingIcon
+  },
+  {
+    text: 'Education',
+    icon: LibraryBooksIcon
+  },
+  {
+    text: 'Pets',
+    icon: PetsIcon
+  },
+  {
+    text: 'Wifi & Tech',
+    icon: WifiIcon
+  },
+  {
+    text: 'Work',
+    icon: WorkIcon
+  },
+  {
+    text: 'Social Services',
+    icon: InfoIcon
+  }
+];
+
+export const HomeComponent = () => (
   <Grid container className="home" spacing={16} justify="space-evenly">
-    <Grid component={Search} item xs={12} />
+    <Grid item xs={12}>
+      <Grid container className="home" spacing={16} justify="space-evenly">
+        <Grid component={Search} item xs={12} />
+      </Grid>
+    </Grid>
+    <Grid item xs={12}>
+      <Grid container spacing={16} justify="space-evenly">
+        {HomeButtonsMajor.map((button, index) => {
+          return (
+            <Grid item xs={4} key={button.text}>
+              <Grid container direction="column" align="center">
+                <IconButton color="primary">
+                  <button.icon />
+                  {button.text}
+                </IconButton>
+              </Grid>
+            </Grid>
+          );
+        })}
+      </Grid>
+    </Grid>
+    <Grid item xs={12}>
+      <Grid container spacing={16} justify="space-evenly">
+        {HomeButtonsMinor.map(button => {
+          return (
+            <Grid item xs={4} key={button.text}>
+              <Grid container direction="column" align="center">
+                <IconButton color="secondary">
+                  <button.icon />
+                  {button.text}
+                </IconButton>
+              </Grid>
+            </Grid>
+          );
+        })}
+      </Grid>
+    </Grid>
   </Grid>
 );
+
+const Home = () => <HomeComponent />;
 
 export default Home;
--- a/src/components/Home.jsx
+++ b/src/components/Home.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
+import { Grid } from '@material-ui/core';
+import Search from './Search';
 
 const Home = () => (
-  <div className="home">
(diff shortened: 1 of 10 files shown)


src/components/Home.test.js
```diff
@@ -1,17 +1,69 @@
...
+    HomeButtonsMajor.forEach(button => {
+      let matchingButtons = wrapper
+        .findWhere(child => {
+          return (
+            child.type() === 'IconButton' &&
+            child.text() === button.text &&
+            child.find(`${button.icon}`).length === 1
+          );
+        })
+        .map(iconButton => iconButton.text());
```
jacobvenable: I also figured out how to return a more significant failure when using [Enzyme's findWhere](https://airbnb.io/enzyme/docs/api/ShallowWrapper/findWhere.html) by mapping the text of the returned button to an array and checking the values of that text instead.

src/components/Home.test.js
```diff
@@ -1,17 +1,69 @@
...
+      expect(matchingButtons).toContain(button.text);
```
jacobvenable: We'll text expect to find the current button's text within that returned array and display missing button's text on failure.

src/components/Home.test.js
```diff
@@ -1,17 +1,78 @@
...
+    );
   });
   it('renders one layout component as the search component', () => {
     expect(
       wrapper.findWhere(n => n.prop('component') === 'Search').length
     ).toBe(1);
   });
+  it('renders all major buttons and only once', () => {
+    HomeButtonsMajor.forEach(button => {
+      let matchingButtons = wrapper
```
gordonhgraham: rather pedantic but, if `matchingButtons` is never reassigned use `const` instead of `let`

jacobvenable: That's in a forEach and reassigned every loop, but it would make sense to define the variable outside of the loop.

src/components/Home.test.js
```diff
@@ -1,17 +1,78 @@
...
+      expect(matchingButtons).toContain(button.text);
+      expect(matchingButtons.length).toBe(1);
+    });
+  });
+  it('renders all minor buttons and only once', () => {
+    HomeButtonsMinor.forEach(button => {
+      let matchingButtons = wrapper
+        .findWhere(child => {
+          return (
+            child.type() === 'Button' &&
```
gordonhgraham: to match the button this returned function must evaluate to true, correct? is checking all of these properties on `child` necessary? I'm not super familiar with `findWhere`, just seems like overkill

jacobvenable: This is just making sure that the compared element is a Button component and then dives into more specific properties based on if that is a true base.

I could technically do `wrapper.find('Button').findWhere(child => {...})` and accomplish the same thing.

The reason I think it's necessary to test all of these props is that I want to make sure that each button prop has the correct value assigned to it. So I in the props below, I'm testing:

1. It's a <Button/>
2. It contains the correct icon
3. It contains a `component` prop that is a <Link/> component.
4. It contains a `to` prop equal to the path specified.

Maybe each one of these could be their own test?

src/components/Home.test.js
rhinodavid: a new context gets created with each function call, so you can use `const`

rhinodavid:
```suggestion
      const matchingButtons = wrapper;
```

src/components/Home.test.js
```diff
@@ -1,17 +1,78 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import Home from './Home';
+import { HomeComponent, HomeButtonsMajor, HomeButtonsMinor } from './Home';
+import { Link } from 'react-router-dom';
+Link.displayName = 'Link';
+jest.mock('@material-ui/core/Button', () => 'Button');
```
rhinodavid: We should find and/or make something that does all these material mocks with one import

rhinodavid: Not for this PR but just in general

jacobvenable: @rhinodavid could you make an issue for that so we remember to address it?

src/components/Home.test.js
gordonhgraham: Yeah, i'd pull them out to their own tests. typically we try to keep a single expectation with a single test. I do see your reasoning here though.

src/components/Home.jsx
```diff
@@ -1,11 +1,128 @@
...
+                </Button>
+              </Grid>
+            </Grid>
+          );
+        })}
+      </Grid>
+    </Grid>
+    <Grid item xs={12}>
+      <Grid container spacing={16} justify="space-evenly">
+        {HomeButtonsMinor.map(button => {
```
gordonhgraham: this looks almost identical to line 93. Maybe pull it out as a helper function `renderButtons()` This time i'd suggest it because it would be more readable, it would help keep the code DRY, but i'm less worried about that.

src/components/Home.test.js
gordonhgraham: check the jest docs on mocking node_modules., i think you could either set up a mock's file and it'll generate mocks for you, or there is a factory arg to the jest.mock function. not sure, but that may help.

src/components/Home.jsx
```diff
@@ -1,11 +1,128 @@
...
+import HomeIcon from '@material-ui/icons/Home';
+import HotTubIcon from '@material-ui/icons/HotTub';
+import BusIcon from '@material-ui/icons/DirectionsBus';
+import GroceryStoreIcon from '@material-ui/icons/LocalGroceryStore';
+import HealingIcon from '@material-ui/icons/Healing';
+import LibraryBooksIcon from '@material-ui/icons/LibraryBooks';
+import PetsIcon from '@material-ui/icons/Pets';
+import WifiIcon from '@material-ui/icons/Wifi';
+import WorkIcon from '@material-ui/icons/Work';
+import InfoIcon from '@material-ui/icons/Info';
```
gordonhgraham: you could probably import these into an index file then require them from there...Like an interface. if you ever are required to switch away from material-ui icon library you'll find it challenging to find/replace every import.

gordonhgraham: do you require some of these icons elsewhere in the codebase?

jacobvenable: Ya, good call. I'll split these out.

src/components/Home.jsx
rhinodavid: Yeah fwiw, I normally inline until I have to copy-paste similar stuff 3ish times unless there's some big readability issue.

rhinodavid:
Any reason why these PRs (this + the carosel) are PRs into other branches (instead of `master`)? Feels a little odd.

jacobvenable:
This is just to break the branches into something that a bit easier to review. This won't be the usual process starting next week. Just something I'm doing for the sake of practicing TDD.

rhinodavid:
> This is just to break the branches into something that a bit easier to review. This won't be the usual process starting next week. Just something I'm doing for the sake of practicing TDD.

@jacobvenable I was thinking about this a little more. I'm definitely for smaller, bite-sized PRs, so I think we should all keep that up. Seems like for this though we'll have to conduct another review once the PR into `master` is created. The way I would expect this to be handled is to just make the PRs into `master`, then just rebase and start a new branch for the next phase of the feature. You'll have multiple PRs for a single issue, but I don't see that as a problem.

jacobvenable:
> > This is just to break the branches into something that a bit easier to review. This won't be the usual process starting next week. Just something I'm doing for the sake of practicing TDD.
> 
> @jacobvenable I was thinking about this a little more. I'm definitely for smaller, bite-sized PRs, so I think we should all keep that up. Seems like for this though we'll have to conduct another review once the PR into `master` is created. The way I would expect this to be handled is to just make the PRs into `master`, then just rebase and start a new branch for the next phase of the feature. You'll have multiple PRs for a single issue, but I don't see that as a problem.

Ya I think that makes sense. I think that after a certain period, if a PR on a sub-branch doesn't get a review and is blocking progress, we should still go ahead and merge and then cover anything missed during the PR into the master branch.

jacobvenable:
@rhinodavid see any other issues? If not, I'm going to go ahead and merge. I'll work on converting items to TS in a separate PR.

rhinodavid:
> @rhinodavid see any other issues? If not, I'm going to go ahead and merge. I'll work on converting items to TS in a separate PR.

Yeah go for it

Issue #29, opened 2019-03-02
[Listing Page] As a person experiencing homelessness, when I hear about a service I need to know some basic information so that I can determine if using the resource is feasible for me.


amaliebarras:
![listing.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/565201200fec5f3e00d1ff86/125516c66e82c728ace21e0d46e49d89378b8daecda5fd53b652fa912a5265296b51a07c23d5bb2bef1b7d550d0e4eac405d0944f8ac892ba3a42062c80d55b98e30063e937f03a6aa8317b500bf258a4a81b0bb8d44c9ed3a37aa.png)

## Info Displayed on a Listing Page
- existing information:
   - [ ] Resource Name
   - [ ] Address
   - [ ] who it is for
   - [ ] what they offer
   - [ ] Full description
   - [ ] website
   - [ ] phone number
   - [ ] whether they are open or closed
   - [ ] whether user has favorited it or not
   - [ ] hours of operation
- missing information
   - [ ] image of the resource (optional)
- sub-components to be built
   - [ ] carousel for multiple images
   - [ ] map embed with #40

jacobvenable:
Since this is an individual listing, should the category bar still be located on the page? @amaliebarras

amaliebarras:
You're right, that's a little weird. I think that could go away. Optionally, we could add navigation back to the listings, but also I think users would likely just use the back button on their browser to get back.

jacobvenable:
Closing in favor of the smaller issues left with this page.

PR #46, opened 2019-03-29
Add/listing page/carousel

jacobvenable:
This PR does not directly address an issue.

## What does this PR do?

Adds a `<Carousel/>` component that will eventually be used for the image slider found on #29.

### DIFF
--- a/src/components/Carousel.jsx
+++ b/src/components/Carousel.jsx
@@ -1,0 +1,13 @@
+import React from 'react';
+
+const Carousel = props => {
+  return (
+    <div>
+      {props.images.map(image => {
+        return <img src={image.url} alt={image.description} />;
+      })}
+    </div>
+  );
+};
+
+export default Carousel;
--- a/src/components/Carousel.jsx
+++ b/src/components/Carousel.jsx
@@ -2,11 +2,11 @@ import React from 'react';
 
 const Carousel = props => {
   return typeof props.images === 'object' && props.images.length ? (
-    <div>
-      {props.images.map(image => {
-        return <img key={image.id} src={image.url} alt={image.description} />;
-      })}
-    </div>
+    <>
+      {props.images.map(image => (
+        <img key={image.id} src={image.url} alt={image.description} />
+      ))}
+    </>
   ) : null;
 };
 
--- a/src/components/Carousel.jsx
+++ b/src/components/Carousel.jsx
@@ -1,13 +1,13 @@
 import React from 'react';
 
 const Carousel = props => {
-  return (
+  return typeof props.images === 'object' && props.images.length ? (
     <div>
       {props.images.map(image => {
         return <img src={image.url} alt={image.description} />;
       })}
     </div>
-  );
+  ) : null;
 };
 
 export default Carousel;
--- a/src/components/Carousel.jsx
+++ b/src/components/Carousel.jsx
@@ -4,7 +4,7 @@ const Carousel = props => {
   return typeof props.images === 'object' && props.images.length ? (
     <div>
       {props.images.map(image => {
-        return <img src={image.url} alt={image.description} />;
+        return <img key={image.id} src={image.url} alt={image.description} />;
       })}
     </div>
   ) : null;
--- a/src/components/Carousel.test.js
+++ b/src/components/Carousel.test.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Carousel from './Carousel';
+
+describe('<Carousel/>', () => {
+  const props = {
+    images: [
+      {
+        url: 'https://placeholder.com/480x200.png',
+        description: 'brief image description here'
+      },
+      {
+        url: 'https://placeholder.com/480x250.png',
+        description: 'brief image description here'
+      }
+    ]
+  };
+  const wrapper = shallow(<Carousel {...props} />);
+  it('renders all images and only once', () => {
+    let renderedImages;
+    props.images.forEach(image => {
+      renderedImages = wrapper
+        .findWhere(child => {
+          return (
+            child.type() === 'img' &&
+            child.prop('src') === image.url &&
+            child.prop('alt') === image.description
+          );
+        })
+        .map(renderedImage => renderedImage.prop('src'));
+      expect(renderedImages).toContain(image.url);
+      expect(renderedImages.length).toBe(1);
+    });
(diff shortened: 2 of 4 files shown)


src/components/Carousel.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const Carousel = props => {
+  return typeof props.images === 'object' && props.images.length ? (
```
rhinodavid: Currently looking at implementing typescript for the project which will help with some of this

src/components/Carousel.jsx
```diff
@@ -0,0 +1,13 @@
...
+    <div>
```
rhinodavid: If we don't actually need a div here (which we should avoid) you can use [`React.Fragment`](https://reactjs.org/docs/fragments.html) (also [see shorthand](https://reactjs.org/docs/fragments.html#short-syntax) -- not sure if we can use it with our build system yet)

src/components/Carousel.jsx
```diff
@@ -0,0 +1,13 @@
...
+      {props.images.map(image => {
```
rhinodavid:
```suggestion
           {props.images.map(image => 
               <img key={image.id} src={image.url} alt={image.description} />
      )}
```

Equivalent but feels so good

rhinodavid:
```
      {props.images.map(image => 
         <img key={image.id} src={image.url} alt={image.description} />
      )}
```

equivalent but feels so good

src/components/Resource.jsx
```diff
@@ -1,81 +1,96 @@
...
+    }
+  ]
 };
 
-export const ResourceComponent = () => {
-  return <div />;
+const Resource = props => {
+  const { images } = singleResourceApiReturn;
+  return (
+    <div>
```
rhinodavid: Same comment about Fragments

jacobvenable:
@rhinodavid other than the TS comment, addressed comments in the latest commit. Thanks!

src/components/Resource.jsx
```diff
@@ -1,81 +1,96 @@
...
+    }
+  ]
 };
 
-export const ResourceComponent = () => {
-  return <div />;
+const Resource = props => {
+  const { images } = singleResourceApiReturn;
+  return (
+    <>
```
rhinodavid: Actually on second thought I don't think we even need the Fragment here. Should just be able to do return(<Carousel images={images} />)

src/components/Carousel.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const Carousel = props => {
+  return typeof props.images === 'object' && props.images.length ? (
+    <>
```
gordonhgraham: typo? new syntax i'm unaware of? looks like a fragment...?

src/components/Carousel.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+describe('<Carousel/>', () => {
+  let wrapper, props;
+
+  beforeEach(() => {
+    props = {
+      images: [
+        {
+          id: '123yvu1231bv23',
+          url: 'https://placeholder.com/480x200.png',
```
gordonhgraham: may i recommend [Fill Murray](https://www.fillmurray.com/)

gordonhgraham: 😹

src/components/Carousel.jsx
jacobvenable: yep, that's just a short version of a fragment:
https://reactjs.org/docs/fragments.html#short-syntax

rhinodavid: see https://reactjs.org/docs/fragments.html#short-syntax

motivation is we don't want to put extra nodes in the dom tree (`<div>`s) if we don't actually need them semantically.

src/components/Resource.jsx
jacobvenable: eventually we will need a fragment, there's just only one component being returned at the moment. The next PR will most likely add another sibling of `<Carousel>`

src/components/Carousel.test.js
```diff
@@ -0,0 +1,59 @@
...
+        }
+      ]
+    };
+    wrapper = shallow(<Carousel {...props} />);
+  });
+  it('renders all images and only once', () => {
+    let renderedImages;
+    props.images.forEach(image => {
+      renderedImages = wrapper
+        .findWhere(child => {
```
gordonhgraham: i'm not crazy about this... rather tough to read. i think it's the chainign of all these methods... `.forEach().findWhere().map()` i'll try to find you an example of another way to do it or Maybe we can pair on another solution.

src/components/Carousel.test.js
```diff
@@ -0,0 +1,59 @@
...
+            child.prop('src') === image.url &&
+            child.prop('alt') === image.description
+          );
+        })
+        .map(renderedImage => renderedImage.prop('src'));
+      expect(renderedImages).toContain(image.url);
+      expect(renderedImages.length).toBe(1);
+    });
+  });
+  it("doesn't render if props are not provided", () => {
```
gordonhgraham: take a look at [prop types](https://reactjs.org/docs/typechecking-with-proptypes.html) That could eliminate two of these tests, checking if props are provided and checking if they're the correctd type.

src/components/Carousel.jsx
gordonhgraham: 👏  thanks guys

src/components/Resource.jsx
rhinodavid: Build it when you need it

src/components/Carousel.test.js
rhinodavid: https://github.com/CodeForBoulder/upswyng/pull/47

sell proptypes >> buy typescript

src/components/Resource.jsx
jacobvenable: You're kill'n me smalls. Updated in last commit.

src/components/Carousel.test.js
jacobvenable: 🤣 I had no idea this existed. I'll have to keep that in mind for next placeholders.

src/components/Carousel.test.js
jacobvenable: I'll leave this as is until we decide which direction to go. We'll discuss on #47

PR #47, opened 2019-03-30
Add TypeScript to the project

rhinodavid:
## What does this PR do?

Adds typescript to upswyng

## Discussion

I've mentioned this before and I'm submitting this PR to continue the discussion about wether we want to commit to this. I'm certainly for it, but I can see good arguments for and against.

Against: This is intimidating to some people and could constitute a barrier to entry for new contributors. It also might slow down development on a micro scale. Fighting the typechecker is real, and definitely _feels_ less flexible to work with. It's also another thing to learn for contributors new to React, et. al.

For: Typescript and similar systems are becoming standard for large and/or production grade projects, and this is a good place to get familiar with it for all of us. It also provides some "guardrails" as people move in and out of the project to keep everyone on the same page. VSCode tooling is fantastic out of the box for TS, and a type system helps enormously to self-document the code we write. Ultimately, I believe a type system will increase dev velocity because it protects against so many gotchas inherent in Javascript. Skeptics should checkout the branch and check it out.

Autoformatters and type checkers give me life


jacobvenable:
I wouldn't be against adding TS, especially if we do it sooner rather than later. While I haven't worked with it before, I see the benefit of being able to declare types as well as implementing a bit of a more object-oriented approach.

I guess what I'm saying is that I'm open to it but neutral. Thoughts @fredmon3 ?

PR #49, opened 2019-04-18
Add/google map #40

### LINKED ISSUE #40: Create Google Map component
Google Maps are used in two separate areas:

A. in the "map view" when viewing a list of services of a certain category on a resource list page #28 
B. on a specific resource page detailing the the location of the resource as well as showing a bus route to the resource #29 

Based on the two separate use cases, this will need to be able to meet the following conditions:

- [ ] accept input of one or more locations to create pins on a map
- [ ] accept a way of showing/hiding directions to an individual location based on the user's current location (if provided)

jmarsto:
This PR closes #40 

## What does this PR do?
Creates a Google Map component, rendering `resources` passed to it as props.
Component needs an API_KEY, I was using mine while working on it.  Happy to share, but they are free and easy to procure your own.
Component expects props such as these:
```no-highlight
const resources = [
  {
    name: "Resource 1",
    category: "Shelter",
    address: "23 Pearl St.",
    audience: "injured",
    services: ["shower", "beds"],
    hours: {},
    lat: 40.0208552,
    lng: -105.267735
  },
  {
    name: "Resource 2",
    category: "Food",
    address: "2280 Junction Pl, Boulder, CO 80301",
    audience: "injured",
    services: ["food", "bathroom"],
    hours: {},
    lat: 40.0242135,
    lng: -105.2518
  },
]

// passed as currentUser
const user = {
  name: "user1",
  location: {
    lat: 40.02,
    lng: -105.2
  }
}
```

Apologies to the TypeScript gods for not abiding by their ways - but I figured I'd get it working and hopefully we can have a meeting of the minds and share some TypeScript knowledge? or I can wade deeper into the docs.  All feedback welcome, I am only a year into web development and eager to learn and to be more helpful than harmful!


src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+              lat: resource.lat,
+              lng: resource.lng
+            },
+            travelMode: 'TRANSIT'
+          },
+          function(response: any, status: any) {
+            if (status === 'OK') {
+              directionsDisplay.setDirections(response);
+            } else {
+              window.alert('Directions request failed due to ' + status);
```
jacobvenable: Let's `console.log` this for now.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+        title: resource.name,
+        position: {
+          lat: resource.lat,
+          lng: resource.lng
+        }
+      });
+
+      let infoWindow = new maps.InfoWindow({
+        content:
+          '<div><b>' +
```
jacobvenable: Any idea if this can accept jsx? If so, maybe a separate method could return the jsx for an info window.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+    let directionsService = new maps.DirectionsService();
+    let directionsDisplay = new maps.DirectionsRenderer();
+    this.setState({
+      map: map,
+      maps: maps,
+      directionsService: directionsService,
+      directionsDisplay: directionsDisplay
+    });
+
+    this.createMarkers(this.props.resources, map, maps);
```
jacobvenable: Since resources exists in props do we need to pass to createMarkers?

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+  };
+
+  render() {
+    const firstResource = this.props.resources[0];
```
jacobvenable: Could destructure this as well:
```javascript
const { name, lat, lng } = this.props.resources[0]
```
Then any use of `firstResource.someProperty` below can just use `someProperty`.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const toggleDirectionsOnClick = () => {
```
jacobvenable: I'd be alright with this anonymous function existing in the onClick attribute rather than in a separate variable. Just a matter of opinion though so it's good either way :P

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    return (
+      // Map must have height/width defined - but manipulate as necessary
+      <div style={{ height: '60vh', width: '100%' }}>
+        <h1>Map</h1>
+        <button onClick={toggleDirectionsOnClick}>
+          Show/Hide Directions to {firstResource.name}
+        </button>
+        <GoogleMapReact
+          bootstrapURLKeys={{ key: '<API_KEY>' }}
```
jacobvenable: While I haven't tried it myself, I think we can create a local environment variable for this to make updating API keys easier. You could replace `'<API_KEY>'` with `{REACT_APP_API_KEY_GOOGLE_MAPS}` and then add that variable to your own [process.env](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables#what-other-env-files-can-be-used) that sits at the top level of this project.

I'm thinking for local dev, it would be best to use our own API keys and then have our centralized one on the live site.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+    };
+    let center = {
+      lat: firstResource.lat,
+      lng: firstResource.lng
+    };
+
+    return (
+      // Map must have height/width defined - but manipulate as necessary
+      <div style={{ height: '60vh', width: '100%' }}>
+        <h1>Map</h1>
```
jacobvenable: For now, update to `<h2>` since there should be only one `<h1>` per page.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+          '<b>For</b>: ' +
+          resource.audience +
+          '<br>' +
+          '<b>Offers</b>: ' +
+          resource.services.join(', ') +
+          '</div>'
+          // will need to include if its open/hours (unsure of how resource.hours will be formatted)
+      });
+
+      marker.addListener('mouseover', function() {
```
jacobvenable: Let's add an extra listener for `focus` as well.

src/components/Map.tsx
```diff
@@ -0,0 +1,145 @@
...
+        infoWindow.open(map, marker);
+      });
+
+      marker.addListener('mouseout', function() {
```
jacobvenable: Add another listener for `blur` too.

src/components/Map.tsx
jmarsto: From what I can tell, it only accepts strings - can we stringify JSX?

src/components/Map.tsx
jmarsto: Sounds like a good approach.  I'm struggling to set that up - I've added it to my .env and I've tried running `require('dotenv').config()` before calling on it, but I think I'm missing something.

.gitignore
```diff
@@ -13,6 +13,7 @@
 
 # misc
 .DS_Store
+.env
```
jacobvenable: I believe you could just use an `.env.local` file. No need to have the separate `.env`.
Update: Saw your comment further down and I'll look into trying to use my own API key locally and get back to you.

src/components/Map.tsx
jacobvenable: As far as I know, not without some kind of external library, but I don't know if it would be worth it at that point. I'm going to suggest adding a TODO comment to revisit this in the future.

src/components/Map.tsx
```diff
@@ -0,0 +1,163 @@
...
+          '<br>' +
+          resource.address +
+          '<br><br>' +
+          '<b>For</b>: ' +
+          resource.audience +
+          '<br>' +
+          '<b>Offers</b>: ' +
+          resource.services.join(', ') +
+          '</div>'
+        // will need to include if its open/hours (unsure of how resource.hours will be formatted)
```
jacobvenable:
```suggestion
        // TODO: Include if its open/hours (unsure of how resource.hours will be formatted)
        // TODO: Use a custom component instead of an infoWindow
```

src/components/Map.tsx
jacobvenable: I'll pull the branch down in a bit to see if I can get it working with my own API key.

src/components/Map.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const centerOnFirstResource = {
+      lat: lat,
+      lng: lng
+    };
+
+    return (
+      // Map must have height/width defined - but manipulate as necessary
+      <div style={{ height: '60vh', width: '100%' }}>
+        <h2>Map</h2>
+        <button onClick={() => this.toggleDirections(firstResource)}>
```
jacobvenable:
```suggestion
        <button onClick={this.toggleDirections}>
```
No need to send first resource to toggleDirections since you can grab it in `toggleDirections()` from props.

src/components/Map.tsx
```diff
@@ -0,0 +1,163 @@
...
+          lat: resourceAsDestination.lat,
+          lng: resourceAsDestination.lng
+        },
+        travelMode: 'TRANSIT'
+      },
+      function(response: any, status: any) {
+        if (status === 'OK') {
+          directionsDisplay.setDirections(response);
+        } else {
+          console.log('Directions request failed due to ' + status);
```
jacobvenable:
```suggestion
          console.log(`Directions request failed due to ${status}`);
```
Just a personal preference on this 😛

src/components/Map.tsx
jmarsto: What if we want to reuse toggleDirections for something other than our first resource?

jacobvenable: Ya, then we would want to think about possibly passing the resource in that manner. Since we're just using the first resource for now, it's just a question of whether we hardcode toggling directions to the first resource in the `render` method vs. in `toggleDirections()`. I'd personally rather have that hardcode exist in toggleDirections but both approaches work.

PR #57, opened 2019-05-18
Feat/filter data boulder

fredmon3:
## What does this PR do?
Adds filtering to the data pipeline.


data_pipeline/filter.js
```diff
@@ -0,0 +1,40 @@
...
+  latHi: 40.260222,
+  lngLow: -105.690812,
+  lngHi: -105.055875
+};
+
+let charity = JSON.parse(
+  fs.readFileSync('data_pipeline/strapped/charity.json')
+);
+
+//TODO: filter
```
jacobvenable: Is this TODO still needed?

fredmon3: Nope. Removed.

Issue #48, opened 2019-04-01
Find and implement a linter we like


rhinodavid:
Prettier is great for formatting. Typescript is great for typing. Add a linter to enforce style issues like Upper/Lower case for vars, etc.

jacobvenable:
How about eslint? Maybe use airbnb rules as a base.

jacobvenable:
Closing because this has been completed 🎉

PR #59, opened 2019-05-24
🏷️ Implement es-lint checks and fixed errors

rhinodavid:
This PR makes progress on #48. I would also like to add a section to the `README` about how to handle type/lint errors.

## What does this PR do?

- Add and configure es-lint to the project, along with TypeScript, React, and Prettier rules.
- Modify lint script to use prettier only on files which es-lint can't handle.
- Fix some errors that occur when adding new linter requirements.


package.json
```diff
@@ -5,38 +5,36 @@
...
     "build": "react-scripts build",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject",
-    "lint": "prettier \"src/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
+    "ci-lint": "prettier --check --loglevel error '{data_pipeline,.github,src,.vscode}/**/*.{json,css,scss,md}' && tsc --noEmit && eslint 'src/**/*.{js,jsx,ts,tsx}' --quiet",
     "datapipe": "npm run strapped && npm run filter",
+    "eject": "react-scripts eject",
     "filter": "node data_pipeline/filter.js",
-    "strapped": "npm run strapped-charity && npm run strapped-crisisline",
+    "lint": "prettier --write --loglevel warn '{data_pipeline,.github,src,.vscode}/**/*.{json,css,scss,md}' && tsc --noEmit && eslint 'src/**/*.{js,jsx,ts,tsx}' --quiet --fix",
```
jacobvenable: Does this script still exist to automatically fix problems?

package.json
```diff
@@ -45,11 +43,17 @@
...
     "enzyme-adapter-react-16": "^1.11.2",
+    "enzyme": "^3.9.0",
+    "eslint-config-prettier": "^4.3.0",
+    "eslint-plugin-css-modules": "^2.11.0",
+    "eslint-plugin-prettier": "^3.1.0",
+    "eslint-plugin-react": "^7.13.0",
     "husky": "^1.3.1",
     "lint-staged": "^8.1.4",
-    "prettier": "1.16.4"
+    "prettier": "^1.17.1"
```
jacobvenable: More of a question than a concern but is there a reason to unpin this?

.vscode/settings.json
```diff
@@ -1,4 +1,24 @@
 {
-    "jest.pathToJest": "npm run test -- --coverage",
-    "editor.formatOnSave": true
-}
\ No newline at end of file
+  "jest.pathToJest": "npm run test -- --coverage",
+  "editor.formatOnSave": true,
```
jacobvenable: since we're disabling this on the main file types, should we just go ahead and turn this off?

package.json
rhinodavid: Ummm no I just copy pastaed from a tutorial I think

package.json
rhinodavid: yes! eslint (with prettier configs) will fix js/ts files, then prettier will do the rest

.vscode/settings.json
rhinodavid: Ummm I don't think so. The goal of all the config is to have only 1 autoformatter run. So the `eslint.autoFixOnSave` will run and fix the kinds that have `editor.formatOnSave` turned off.

rhinodavid: we still want `editor.formatOnSave` on, for instance, `css` or `json` files

PR #65, opened 2019-06-05
Listing Page #1: Wrap <Resource/> Component

jacobvenable:

## What does this PR do?

- creates a `<ResourceContainer/>` that
  - grabs an ID search parameter from the URL
  - feeds that id to the `withData` component along with `<Resource/>` and returns the result
- updates `<Resource/>`
  - remove the use of `id`
  - updates the `resource` prop to be optional
- moves both of those components into their own folder
- creates an index file to export `<ResourceContainer/>` by default
- modifies/adds tests for `<Resource/>` and `<ResourceContainer/>`

## What's next?

- I'll be implementing this new container in `<App/>` so we can see some data rendered.


jacobvenable:
I'm going to refactor this and create another PR.

PR #67, opened 2019-06-08
Add getSearchParam utility

jacobvenable:
This PR does not address an issue.

## What does this PR do?

Adds a `getSearchParam` utility that can be used to grab the value of a URL search parameter. If the parameter does not exist or has no value, the utility will return `null`.

This utility will eventually be used in a future PR to grab a resource's ID from the URL.

### DIFF
--- a/src/utils/searchParams.ts
+++ b/src/utils/searchParams.ts
@@ -1,0 +1,5 @@
+export const getSearchParamVal = (paramName: string) => {
+  const allSearchParams = new URLSearchParams(location.search);
+
+  return allSearchParams.get(paramName) || null;
+};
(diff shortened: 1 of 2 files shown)


src/utils/searchParams.ts
```diff
@@ -0,0 +1,5 @@
+export const getSearchParamVal = (paramName: string) => {
```
rhinodavid:
```suggestion
export const getSearchParamVal = (paramName: string) : string | null  => {
```

jacobvenable: [updated](https://github.com/CodeForBoulder/upswyng/pull/67/commits/e7086b572034da73f939206da2fed37d0f339e98)

jacobvenable:
I also updated the utility function to allow returning an empty string for parameters without value. Realized that empty strings are falsey anyways.

PR #70, opened 2019-06-12
Create combine npm script

jacobvenable:

## What does this PR do?

- This creates a new `combine` npm script that combines the `data_pipeline/charity.json` and `data_pipeline/crisisline.json` into `all.json`.
- Adds `data_pipeline/all.json` (mentioned above) containing the database info.
- This also adds the `combine` script to also be run in the `datapipe` script.

## Why though?

Having a JSON file containing the DB will make creating DBs for local development a bit easier.

fusion


data_pipeline/all.json
```diff
@@ -0,0 +1,5138 @@
+{
```
rhinodavid: My general thought on this is that for this project "buildable" files like this probably shouldn't get committed. If someone changes a file on which this depends but doesn't run the `combine` script, then this file is now out of date. It also adds to the repo size unnecessarily.

I feel the better approach would be just to write the script, put the output file name in the `.gitignore`, then in those instructions you made just say like 'hey run the script and then upload the output file to firebase'.

package.json
```diff
@@ -27,7 +27,8 @@
   "scripts": {
     "build": "react-scripts build",
     "ci-lint": "prettier --check --loglevel error '{data_pipeline,.github,src,.vscode}/**/*.{json,css,scss,md}' && tsc --noEmit && eslint 'src/**/*.{js,jsx,ts,tsx}' --quiet",
-    "datapipe": "npm run strapped && npm run filter",
+    "combine": "node data_pipeline/combine.js",
+    "datapipe": "npm run strapped && npm run filter && npm run combine",
```
rhinodavid: do we want to just go ahead and remove the `strapped` references?

data_pipeline/all.json
jacobvenable: I believe the original reason to add it to the repo was to have some kind of version control on that data itself. We'll need to talk with @fredmon3 a bit more on that since we hadn't completely planned out how that would work.

package.json
jacobvenable: I think we should keep it since that command actually goes and grabs data from strapped. Once we're completely independent, we can remove these commands.

data_pipeline/all.json
rhinodavid: Yeah I mean that's fine.. have the two files that combine to make "all" under version control, just not the file that gets built with the two

rhinodavid: Oh wait I see.. this actually does a web request to strapped. ok hold on

data_pipeline/combine.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
+
+const charity = JSON.parse(fs.readFileSync('data_pipeline/charity.json'));
+const hotline = JSON.parse(fs.readFileSync('data_pipeline/crisisline.json'));
+
+const db = {
+  charity,
+  hotline
```
fredmon3: Would it be a good idea to leave this collection as `crisisline` like the original dataset?

jacobvenable: good call. [updated](https://github.com/CodeForBoulder/upswyng/pull/70/commits/a2c1eb564ac7476f7900b6d640d1c46bfd083a43)

data_pipeline/all.json
jacobvenable: [update](https://github.com/CodeForBoulder/upswyng/pull/70/commits/980dd2b3a2895b25d07c59fd7d095de19f8805ca): removed `all.json` and updated gitignore

PR #71, opened 2019-06-12
Add Firebase Setup Instructions to README

jacobvenable:
This PR doesn't close an issue.

_Note: We'll need to merge #70 first, since part of the instructions rely on a file created in that PR._

## What does this PR do?

Adds setup instructions for Firebase and `.env.local` variables for the project.


rhinodavid:
I guess this is okay for now. I was actually just rooting around Firebase tonight trying to get familiar. A few thoughts:
1) We don't need any real time database features and I think if we're sticking with Firebase we should switch to Cloud Firestore (see: https://firebase.google.com/docs/database/rtdb-vs-firestore).
2) It seems like a bit much for anyone working on this to have to go set up their own test database. I have a few ideas for that:
- Set up a separate Firestore DB with test data and just share that DB's credentials with anyone who wants to use it for dev
- Just use the real database. The app doesn't do any writes so who cares.
- (Prob my favorite, but the most challenging) Modify the Firebase setup file so that if the person spins up the app without specifying those env variables a Firebase fake gets loaded that just serves some hardcoded data.
3) We can commit the env variables.
> Note: The Firebase config object contains unique, but non-secret identifiers for your Firebase project. https://firebase.google.com/docs/web/setup#config-object

jacobvenable:
> 1. We don't need any real time database features and I think if we're sticking with Firebase we should switch to Cloud Firestore (see: https://firebase.google.com/docs/database/rtdb-vs-firestore).

That makes sense to me. I think the only issue is that Strapp'd, where we currently get our data, uses rtdb. We'd need to consider how we would eventually migrate their data as well.

I'm not really about this for our MVP though, so I think it's best to stick with what we already have for now and revisit later.

> 2. It seems like a bit much for anyone working on this to have to go set up their own test database. I have a few ideas for that:

While the directions are deceiving for the setup length, I do agree that it's more than I'd like.

> * Set up a separate Firestore DB with test data and just share that DB's credentials with anyone who wants to use it for dev

I think I had originally worried about doing this way, because it meant contributors would rely on us releasing them per request.

> * Just use the real database. The app doesn't do any writes so who cares.

We haven't setup the final DB that the app will use yet, but the option you have above would work.

> * (Prob my favorite, but the most challenging) Modify the Firebase setup file so that if the person spins up the app without specifying those env variables a Firebase fake gets loaded that just serves some hardcoded data.

Ya, I thought it would be neat to implement Docker to setup something like this with one command. I just didn't want to spend time on trying to figure that out when I could be developing the app.

> 1. We can commit the env variables.
> 
> > Note: The Firebase config object contains unique, but non-secret identifiers for your Firebase project. https://firebase.google.com/docs/web/setup#config-object

That... is a good point. As long as we don't do something wonky with the data via the Firebase console, it shouldn't be a problem for development. Maybe we add those values directly to `env.local.example` and just instruct them to copy it to `env.local`. We'll still need to have a separate example file since the Google Maps API key may need to be dev-unique for local development.

rhinodavid:
> Ya, I thought it would be neat to implement Docker to setup something like this with one command. I just didn't want to spend time on trying to figure that out when I could be developing the app.

I was just thinking about commiting some sample data in the repo and then hijacking the data provider somehow... idk

jacobvenable:
@rhinodavid Made some updates to the README since you last looked and added some public Firebase configuration to the `env.local.sample`. Figured I'd re-request a review 🙏

PR #75, opened 2019-06-19
Update <Search/> to Redirect with Query Parameter

jacobvenable:

## What does this PR do?

Updates the `<Search/>` to:
- to be TypeScript
- to be a class component
- to redirect upon submitting the search


src/components/Search.tsx
```diff
@@ -0,0 +1,69 @@
...
+    e.preventDefault();
+    this.setState({
+      submitted: true
+    });
+  };
+
+  render() {
+    const { submitted, query } = this.state;
+    return (
+      <Fragment>
```
rhinodavid: May be able to use shorthand here `<>`

jacobvenable: [updated](https://github.com/CodeForBoulder/upswyng/pull/75/commits/edfdd51ccba1b9ba9ef73c893c3eb4e8d906be1e) 👍

PR #76, opened 2019-06-23
🧪 Create hook to load a resource

rhinodavid:
## What does this PR do?
This is a React Hooks version of 'withResource' and has the same
ultimate functionality.

dustin hoffman playing the character "Hook" smiling creepily


src/components/__tests__/useResource.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  mockApp = new Mock(); // eslint-disable-line @typescript-eslint/no-use-before-define
+  return {
+    __esModule: true,
+    default: mockApp.initializeApp({}) // eslint-disable-line @typescript-eslint/no-use-before-define
+  };
+});
+
+import useResource from '../useResource';
+
+var mockApp: Mock; // using `var` here so it is hoisted above the mock -- would love a better solution
```
jacobvenable: Unfortunately, I don't have a better solution for ensuring that variables are defined before declaring a mock. Something I have seen kind of work before is [doMock](https://jestjs.io/docs/en/jest-object.html#jestdomockmodulename-factory-options), which allows you to dynamically update mocks. You just have to reimport the module being tested.

So it would be something like:

```js
const mockApp = new Mock();

jest.doMock('../whateverModule', () => ({
  __esModule: true,
  default: mockApp.initializeApp({})
}));

const useResource = require('../useResource'); // using require since import is hoisted (I think)

//kewl tests here
```

This does mean you that you have to use `require` as well which also goes against our current conventions 🙁

src/components/__tests__/useResource.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+it('should set init value', async () => {
+  const { result } = renderHook((path: string) => useResource(path), {
+    initialProps: '/resources/ex-id'
+  });
+  expect(result.current).toBeNull();
+});
+
+it('should update the result when a resource has been recieved', async () => {
```
jacobvenable:
```suggestion
it('should update the result when a resource has been received', async () => {
```

src/components/useResource.tsx
```diff
@@ -0,0 +1,47 @@
...
+ *     <div>
+ *       <h1>{charityData.charityName}</h1>
+ *       <p>{charityData.description}</p>
+ *     </div>
+ *   ) : (
+ *     <Spinner />
+ *   );
+ * };
+ */
+function useResource(dataPath: string): TResource | null {
```
jacobvenable: Not a big deal, but is there a reason to use `function` here instead of an arrow function? Wasn't sure if this was convention I haven't seen yet.

src/components/__tests__/useResource.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { Mock } from 'firebase-nightlight';
```
jacobvenable: Do we want to officially move all tests to this `__tests__` folder? I'm used to having them exist with the components, but usually I follow the convention of also having a folder for component anyways 👍 Either way works as long as we have a standard.

src/components/useResource.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [resource, setResource] = useState(null);
+  useEffect(() => {
+    // Update 'resource' when the Firebase Realtime Database data
+    // under the '/someData' path changes.
```
jacobvenable: I'd be good w/ removing these comments

src/components/useResource.tsx
rhinodavid: hahaha what? writing a function?

src/components/useResource.tsx
rhinodavid: yeah agree

src/components/__tests__/useResource.test.tsx
rhinodavid: Yeah this is all pretty shabby..

src/components/__tests__/useResource.test.tsx
rhinodavid: yeah I was gonna put up a PR to do that.. I think it makes sense as we get more files. I don't have a strong opinion either way.

rhinodavid:
> Few comments and questions but nothing blocking. Before merging, we may want to update this to be called `useResources` since we'll want to use this for grabbing multiple resources as well.

Do you have a list of what fetches we'll make? Do we need a `useResource` and a `useResources`? Maybe there's a better way to organize it?

PR #83, opened 2019-06-28
add initial search page

jacobvenable:
This PR doesn't close an issue.

## What does this PR do?

_**Note**: This should be merged **after** #82. Builds will fail until that is merged and this is rebased_

Adds a basic search page route that is accessed by entering a search term via the `<Search/>` found on the home page.

Styles and elements placed on the page will be updated in a later PR.

The search page takes the search parameter and sends a search request to our development index in Algolia.

Some public env variable values were also added to the `.env.local.example`.

I will find you


src/components/SearchResults.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Link } from 'react-router-dom';
+import algoliaSearch from 'algoliasearch';
+import useSearchResults from './useSearchResults';
+import { SEARCH_PARAM_RESOURCE } from '../constants';
+
+const renderResults = (results: algoliaSearch.Response) => {
+  const { hits } = results;
+  if (hits && hits.length) {
+    const listItems = hits.map(({ charityname, objectID }) => {
+      if (charityname && objectID) {
```
rhinodavid: is this check really needed? Aren't these always defined in `TResource`?

src/components/SearchResults.tsx
```diff
@@ -0,0 +1,41 @@
...
+            >
+              {charityname}
+            </Link>
+          </li>
+        );
+      }
+    });
+    return <ul>{listItems}</ul>;
+  }
+  return null;
```
rhinodavid: So if I'm reading this right, the hook returns `[]` for no hits and `null` if the query is still in progress. I haven't looked at this super closely, but just thinking maybe we want to show some loading indicator for the second case?

src/components/__tests__/SearchResults.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { SEARCH_PARAM_QUERY } from '../../constants';
+
+jest.mock('../useSearchResults');
+jest.mock('react-router-dom', () => ({
+  Link: 'Link'
+}));
+
+describe('<SearchResults/>', () => {
+  const getSubject = () => shallow(<SearchResults />);
+  describe('when it does not receive any search hits', () => {
```
rhinodavid: personal preference would be just the `it` with like 'renders no list for empty search hits` or something. no need to change this.

src/components/__tests__/SearchResults.test.tsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
```
rhinodavid: we should look into snapshots for stuff like this

src/components/SearchResults.tsx
jacobvenable: meh, maybe and maybe not. Although this is actually what comes back from algolia, which isn't an entire `TResource`. I more just wanted to specify that something should only happen if we have these required items. I'm good either way though if you think it should be removed.

src/components/SearchResults.tsx
jacobvenable: The hook returns an `algoliaSearch.Response` object or null when not retrieved.

We will want to show a loading icon if it's still loading but I'm not going to worry about that for this PR. I imagine we'll want to use a material icon for that, but it may be wise to create a component for that to be used throughout the app.

src/components/__tests__/SearchResults.test.tsx
jacobvenable: ya matter of preference really, I like reading the output as:
```
when it does not receive any search hits
  it does not render a list of results
```

src/components/__tests__/SearchResults.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  describe('when it receives at least one search hit', () => {
+    const mockHits = [
+      {
+        charityname: 'Mikes Tots',
+        objectID: '-someRandomId'
+      },
+      {
+        charityname:
+          "Al Harrington's Wacky Waving Inflatable Arm-Flailing Tubeman Emporium and Warehouse",
```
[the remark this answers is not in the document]
jacobvenable: Wacky Waving Inflatable Arm-Flailing Tubeman

src/components/__tests__/SearchResults.test.tsx
jacobvenable: I think snapshots will come in handy once we have a pretty good handle on the final output of components, but until then we'll just get snapshot warnings over and over while we work out the final UI.

src/components/SearchResults.tsx
jacobvenable: I'll go ahead and merge for now, but this will most definitely be revisited when we hammer out how each individual listing appears.

PR #84, opened 2019-07-03
Styled Components for the Home Page

jacobvenable:

## What does this PR do?

This one is a bit of a doozy one. I wanted to see what styling material-ui via styled components would entail, so I hope this will give peeps a good idea of what it could look like.

This is my first time using styled components, so there is certainly a better approach to what I've done here. I don't fully expect this to actually get merged, but to review if this is the way forward.

**I will need to fix some tests to merge if we do think this is the way forward.**

## Cons of Styled Components
- It seems that we'll almost always need to use a duplicate class selector, e.g. `.classname.classname`, to override default material styles. This is done using a [`&&` selector](https://github.com/CodeForBoulder/upswyng/pull/84/files#diff-e2471acc7248081dc1bd135bbe6192ffR16).
- It also looks like some child elements generated by material-ui also need to have a [2 level selector](https://github.com/CodeForBoulder/upswyng/pull/84/files#diff-9ded5085a9b669646645c60c69947b26R16) in order to target them. While this isn't too terrible, it's definitely not ideal.

## Pros
- I'm not sure if any other CSS in JS solution would fix the cons above, other than possibly using style solution provided by material-ui.

The cool thing is I was able to get roughly close to the home page mock, minus a couple elements and some positioning issues.

Screen Shot 2019-07-03 at 8 37 29 AM

experimental


rhinodavid:
Did the TypeScript type checker give you any grief about this? I tried to do `styled(SomeComponent)` the other day and it got angry. I didn't dig too deep though.

src/components/Header.tsx
```diff
@@ -2,28 +2,44 @@ import React from 'react';
...
 
-function Header() {
+const StyledHeader = styled(AppBar)`
+  && {
+    background: none;
+    box-shadow: none;
+    margin-top: ${headerVerticalMargin}px;
+    margin-bottom: ${headerVerticalMargin / 2}px;
+  }
+` as typeof AppBar;
```
rhinodavid: This is probably what I needed to make TS happy.

src/components/Header.tsx
```diff
@@ -2,28 +2,44 @@ import React from 'react';
...
+import { Container } from '../App.styles';
+
+const headerVerticalMargin = 24;
 
-function Header() {
+const StyledHeader = styled(AppBar)`
+  && {
+    background: none;
+    box-shadow: none;
+    margin-top: ${headerVerticalMargin}px;
```
rhinodavid: super duper nit: sort alphabetically

src/components/Search.tsx
```diff
@@ -2,8 +2,44 @@ import React, { Component } from 'react';
...
+    color: ${colors.black};
+  }
+  && > div {
+    margin-top: 0;
+  }
+  label {
+    ${ScreenReaderOnly}
+  }
+  input {
+    padding: 7px 7px 7px 0;
```
rhinodavid: Any reason we're mixing pixels and `em`s? [`rem`](https://www.sitepoint.com/understanding-and-using-rem-units-in-css/) ftw

src/components/Search.tsx
```diff
@@ -2,8 +2,44 @@ import React, { Component } from 'react';
...
+    font-family: ${fonts.openSans};
+    font-size: 1em;
+  }
+  input::placeholder {
+    color: ${colors.black};
+    opacity: 1;
+  }
+` as typeof TextField;
+
+const SearchAdornment = styled(InputAdornment)`
```
rhinodavid: you fancy huh

rhinodavid:
I'm good with this if yall are

src/components/Header.tsx
jacobvenable: Yep that is exactly why that is there 😅

src/components/Header.tsx
jacobvenable: You mean all CSS properties sorted alphabetically?

src/components/Search.tsx
jacobvenable: The pixels were just added to make testing styling easier. I'll create a separate PR that implements a helper function to convert pixel values to rems and implement where necessary.

src/components/Header.tsx
rhinodavid: yeh

jacobvenable: I see the merit in this but it's going to take some getting used to for me 😅 [updated](https://github.com/CodeForBoulder/upswyng/pull/84/commits/ba2c1e8baaf98184857f87596db045d706dd6440)

PR #85, opened 2019-07-07
Fix Algolia Search

jacobvenable:

## What does this PR do?

I noticed that an error was being thrown on `<SearchResults/>` whenever it was unmounted. It turned out that I implemented some unnecessary cleanup on `<useSearchResults/>` by using `algoliaClient.destroy()`, which is only necessary when doing [searches server-side](https://www.npmjs.com/package/algoliasearch#keep-alive).

derp


jacobvenable:
> hero

Screen Shot 2019-07-07 at 9 57 55 PM

That's pro to you.

PR #86, opened 2019-07-08
Landing Page: Update Icons

jacobvenable:
This PR does not close an issue.

## What does this PR do?

- adds the following custom icons
  - banana icon
  - hygiene icon
  - socks icon
- adds/removes icons to the main icons wrapper
- updates the home page buttons to use the proper icons
- fixes button heights not stretching revealed by using custom icons

## How does this PR make you feel? [:link:](http://giphy.com/categories/emotions/)

icons!

PR #87, opened 2019-07-10
Landing Page: add coordinated entry button

### LINKED ISSUE #26: [Landing Page] As a person experiencing homelessness, I need to land on a main page so that I can orient myself in the application.
- menu should say 'menu'
- logo should be ever present inline with the menu 
- search should be prominent on the main page
- coordinated entry button must stay
- more important content (food, shelter) should be more prominent than less important (education, pets)
- Icons could be laid out in blocks instead of circles in order to use space better

jacobvenable:
This PR closes #26 assuming #86 merges first. There may be future modifications, but this will complete the initial version.

## What does this PR do?

- updates the `THomeButton` type to accomodate both `<a/>` and `<Link/>` elements
- adds a door icon
- updates the styled `HomeLink` to output either a `<a/>` or `<Link/>` depending on the present props.
- adds a global style applying `box-sizing:border-box` to all elements.


src/App.styles.ts
```diff
@@ -52,6 +52,9 @@ export const Container = styled(Grid)`
 ` as typeof Grid;
 
 export default createGlobalStyle`
+  * {
```
rhinodavid: thank you

src/components/HomeButtons.tsx
```diff
@@ -130,18 +184,16 @@ const HomeButtons = () => (
...
-              buttonColor={button.color}
-            >
-              {button.text}
-              {button.icon}
-            </HomeButton>
-          </HomeLink>
-        </Grid>
+        <HomeLink {...button} key={button.text}>
+          <HomeButton
+            disableRipple={true}
```
rhinodavid: why?

src/types.ts
```diff
@@ -64,8 +65,10 @@ export interface TSchedule {
 }
 
 export interface THomeButton {
-  text: string;
-  icon: React.ReactElement<SvgIconProps>;
-  to: string;
   color: string;
+  href?: string;
```
rhinodavid: is this just for the links thats open in a new window?

src/components/HomeButtons.tsx
```diff
@@ -22,75 +23,127 @@ const buttons: THomeButton[] = [
...
+    target: '_blank'
   }
 ];
+interface HomeLinkProps extends THomeButton {
+  children: React.ReactElement | React.ReactElement[];
+  key: string;
+}
 
-const HomeLink = styled(Link)`
+const HomeLink = styled((props: HomeLinkProps) => {
```
rhinodavid: my inclination is to break out the "normal" button and the one-off that opens a link

src/components/HomeButtons.tsx
jacobvenable: Users won't be able to see the ripple since they're instantly routed to a different page. Since it doesn't really matter, I just [removed the prop](https://github.com/CodeForBoulder/upswyng/pull/87/commits/a3fdf2968081252088f3a06cb28f054a92fe9639).

src/types.ts
jacobvenable: Yes, but I [separated to two types](https://github.com/CodeForBoulder/upswyng/pull/87/commits/528bc8c126123fbf3e9a90fd9540612e9e192f19).

src/components/HomeButtons.tsx
jacobvenable: [done](https://github.com/CodeForBoulder/upswyng/pull/87/commits/528bc8c126123fbf3e9a90fd9540612e9e192f19). `<HomeButtons/>` was also getting pretty large so I [separated `<HomeLink/>` and `<HomeButton/>` into different modules](https://github.com/CodeForBoulder/upswyng/pull/87/commits/58c0fb885a79a7c77eaf8e66f990d7bfbf0535c9).

src/App.styles.ts
jacobvenable: It was a pleasure.

PR #116, opened 2019-08-15
♻️ Consolidate categories into one module

rhinodavid:
## What does this PR do?
This takes the categories, which were spread across multiple modules/files, and puts them into one place. My hope and prayer is that this makes maintenance a bit easier.


src/App.tsx
```diff
@@ -26,7 +30,7 @@ class App extends Component {
             <Header />
             <Route exact path="/" component={Home} />
             <Route exact path="/shelters" component={Shelters} />
-            <Route exact path="/job-training" component={JobTaining} />
```
rhinodavid: BOOOOOOOOOOOOOOOOO

jacobvenable: ashamed

PR #209, opened 2020-01-03
fix rendering of Weekly schedules

jacobvenable:

## What does this PR do?

Fixes the rendering of weekly schedules in the public section of the app.

example of rendered weekly schedule

_Note: This does not include monthly or annual schedules._

fixed it

### DIFF
--- a/packages/upswyng-web/package.json
+++ b/packages/upswyng-web/package.json
@@ -9,7 +9,7 @@
     "@material-ui/lab": "^4.0.0-alpha.27",
     "@types/algoliasearch": "^3.30.12",
     "@types/enzyme": "^3.9.1",
-    "@types/enzyme-adapter-react-16": "^1.0.5",
+  "@types/enzyme-adapter-react-16": "^1.0.5",
     "@types/google-map-react": "^0.23.11",
     "@types/jest": "^24.0.11",
     "@types/material-ui": "^0.21.6",
--- a/packages/upswyng-web/src/components/Schedule.tsx
+++ b/packages/upswyng-web/src/components/Schedule.tsx
@@ -10,15 +10,19 @@ interface ScheduleProps {
   schedule: TResourceScheduleData;
 }
 
-const days = [
-  "Sunday",
-  "Monday",
-  "Tuesday",
-  "Wednesday",
-  "Thursday",
-  "Friday",
-  "Saturday",
-];
+const days = {
+  sunday: "SU",
+  monday: "MO",
+  tuesday: "TU",
+  wednesday: "WE",
+  thursday: "TH",
+  friday: "FR",
+  saturday: "SA",
+};
+
+function capitalize(x: string): string {
+  return x.charAt(0).toUpperCase() + x.slice(1);
+}
 
 const WeeklyContainer = styled.div`
   display: flex;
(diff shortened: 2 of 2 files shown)


rhinodavid:
Me likey. I'll take a look later today.

packages/upswyng-web/src/components/Schedule.tsx
```diff
@@ -54,78 +54,47 @@ const WeeklyTime = styled.span`
...
+  const dayItemsMap = days.map((day, dayIndex) => {
     return {
       day,
       items: items.filter(
         item =>
           item.recurrenceRule.options.freq === RRule.WEEKLY &&
-          item.recurrenceRule.options.byweekday.includes(
-            (RRule as any)[dayToCode[day as keyof typeof dayToCode]] as number
-          )
+          item.recurrenceRule.options.byweekday.includes(dayIndex)
```
rhinodavid: logic error around hurr

packages/upswyng-web/src/components/Schedule.tsx
```diff
@@ -138,12 +107,12 @@ const Schedule = ({ schedule }: ScheduleProps) => {
     // group together schedule items that have the same comments
     const commentMap = resourceSchedule.getItems().reduce((result, item) => {
       const comment = item.comment || "_no_comment_";
-      result[comment as keyof typeof dayToCode] = [
-        ...(result[comment as keyof typeof dayToCode] || []),
+      result[comment as keyof typeof days] = [
```
rhinodavid: i think you can prob get rid of all this type info if you're just gonna use an array for `days`.. `keyof typeof days` reduces to `number`

rhinodavid: TS may fight you tho

rhinodavid:
Fixed the whole off-by-a-day thing by going back to the RRule code. Still needs a little work though I think.

Screen Shot 2020-01-05 at 4 42 43 PM
Screen Shot 2020-01-05 at 4 42 34 PM

packages/upswyng-web/package.json
```diff
@@ -9,7 +9,7 @@
     "@material-ui/lab": "^4.0.0-alpha.27",
     "@types/algoliasearch": "^3.30.12",
     "@types/enzyme": "^3.9.1",
-    "@types/enzyme-adapter-react-16": "^1.0.5",
+  "@types/enzyme-adapter-react-16": "^1.0.5",
```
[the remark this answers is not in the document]
jacobvenable: Odd, usually prettier fixes these types of things, but I'm not sure why it changed in the first place.

> also already in the root package.json

Our client CI in Netlify only knows of the packages in this file, correct? If so, then that means it needs to be kept, yes?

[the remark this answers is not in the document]
rhinodavid: Um i _think_ it runs 'yarn' at the root so it should do everything..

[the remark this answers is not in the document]
rhinodavid: Also I think the prettier setup may be not doing these json files at the roots

packages/upswyng-web/src/components/Schedule.tsx
rhinodavid: Outdated

jacobvenable:
> I'm good with this minus the little package.json spacing thing.

> I think next we need to look at the design of this page. My gripe is that we're doing columns on mobile, which already isn't ideal and wastes a lot of space when just the heading is on the left. And then with the schedule we further subdivide into 'schedule' '' '' and it just seems v crowded.

Agreed. Let me see if I can put together something a little prettier and make another PR. I'll also working on handling output of other formats.

PR #213, opened 2020-01-05
Add cypress testing to web and server

jollyjerr:
Here is the cleaned-up version of my previous pull request that should play nicely with master :)

I also removed the server E2E testing so that we can start afresh with db mocking and api verification which I will have some questions about on Wednesday!

*Cheers!*


jollyjerr:
I think the builds failed because cypress either timed out downloading - or was looking for localhost:3001 on the web package (I'm not totally sure how you want to integrate these tests in the pipelines so feel free to edit cypress.config)

rhinodavid:
Thanks dude -- let me dig into this tomms

rhinodavid:
Is there a way we can pare down the scope of this. It's a lot now, and its not clear to me right off the bat how to get it all working. Specifically, it requires a working server and working database, and we have no canonical test database setup and the tests don't spin up a server.

Here's what I'd propose:
- Start with one little test. Instead of testing the web app let's do one page; probably the page displaying a resource makes the most sense.
- Mock the server response.
- Once we get some simple assertions working with that, maybe we can look at adding screenshots?

Once we have a test layer which mocks the server, then we can build on that to get to E2E

PR #223, opened 2020-01-07
feat: begin replacing styled-components with MUI styles

jacobvenable:

## What does this PR do?

In preparation for updating the resource hours styles, I'm beginning to replace styled components w/ [Material UI's styling solution](https://material-ui.com/styles/basics/).

Any components touched in this PR needed to be updated due to styles being passed down by the theme provider.

_Note: In updating the `<SearchInput/>`, I did update that to be a functional component w/ a hooks-based approach to use the `makeStyles` hook._

wire mess


packages/upswyng-web/src/components/SearchInput.tsx
```diff
@@ -1,109 +1,71 @@
...
-            }}
-          />
-        ) : null}
-      </>
-    );
-  }
-}
+        </FormControl>
+      </form>
+      {submitted ? (
```
rhinodavid:
```suggestion
      {submitted && (
```

packages/upswyng-web/src/components/SearchInput.tsx
```diff
@@ -1,109 +1,71 @@
...
+        <Redirect
+          to={{
+            pathname: "/search",
+            search: `?${SEARCH_PARAM_QUERY}=${query}`,
+          }}
+        />
+      ) : null}
```
rhinodavid:
```suggestion
      )}
```

packages/upswyng-web/src/components/__tests__/SearchInput.test.tsx
```diff
@@ -3,7 +3,8 @@ import React from "react";
 import SearchInput from "../SearchInput";
 
 jest.mock("react-router/Redirect", () => "Redirect");
-jest.mock("@material-ui/core/TextField", () => "TextField");
+jest.mock("@material-ui/core/Input", () => "Input");
+jest.mock("@material-ui/styles/makeStyles", () => () => () => ({}));
```
rhinodavid: fun

packages/upswyng-web/src/components/SearchInput.tsx
jacobvenable: oh yarp true dat. [updated](https://github.com/CodeForBoulder/upswyng/pull/223/commits/ed401d0053abb88011435090e8c0623b060f8968)

packages/upswyng-web/src/components/SearchInput.tsx
jacobvenable: [upderped](https://github.com/CodeForBoulder/upswyng/pull/223/commits/ed401d0053abb88011435090e8c0623b060f8968)

packages/upswyng-web/src/components/__tests__/SearchInput.test.tsx
jacobvenable: no 😞  I'm going to need to figure out how to manually mock this at a global level so that this doesn't have to happen in each component where `makeStyles` is used.

jacobvenable:
> Why'd you decide to change to MUI? I think styled components is blah but this seems about the same? [Emotion](https://github.com/emotion-js/emotion) ftw.

MUI's solution has been easier to use in the past.

1. We can remove the need for global styling and instead rely on styles being propagated from our customized theme, with little work on our part.
2. Emotion is going to have similar issues with nested styles that styled components does. In MUI's [example using emotion](https://material-ui.com/guides/interoperability/#the-css-prop), you'll see that we need to nest classes in order to target MUI-generated children. With MUI's solution, we can target these children via the built-in classes object provided in each component ([example of the classes object for an MUI input](https://material-ui.com/api/input/#css)). The classes also enables us to target based on conditions (e.g. if a Input is focused, apply these styles to the root or if Input is collapsed, apply these styles)

Both points above mean we can almost completely forget about CSS specificity.

Side note: While this does lean further into MUI, I don't see us getting away from MUI any time soon. Best to stick to an established design system rather than attempt to roll our own (for now).

PR #226, opened 2020-01-09
Mention requiring node version 12.14.x

bepetersn:
## What does this PR do?

Adds a helpful note about an issue that happened while I was setting up the project for the first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ and install the project dependencies by running
 yarn
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,8 @@ and install the project dependencies by running
 yarn
 ```
 
+Note: the above command may fail if your `node` version is not 12.14.x. [nvm](https://itnext.io/nvm-the-easiest-way-to-switch-node-js-environments-on-your-machine-in-a-flash-17babb7d5f1b#d594) is one solution (among others) to manage `node` versions precisely.
```
rhinodavid: ew. n for the win. https://github.com/tj/n

PR #228, opened 2020-01-10
WIP: Docker

jollyjerr:
## What does this PR do?
This PR adds the ability to launch a development mongo DB in a docker container and connect to it from your dev server. I'm marking this as a WIP because I will need help understanding the final goal of this feature and running a script to feed the db with dummy data. 👾

#### Instructions
1. Clone this branch
2. Make sure you have [Docker](https://docs.docker.com/docker-for-mac/) installed
3. Run `docker-compose up` *or `docker-compose up -d` to start as a background service*
4. Add the db .env variables from .env.example on upswyng-server package
5. Start upswyng-server `yarn workspace @upswyng/upswyng-server run start`

*note: you do not need mongodb installed on your machine for this to work - the container will automatically download all it needs*

#### Questions/Concerns
1. How can we reliably populate this database?
2. Currently, the data lives inside the container and will be destroyed if you stop the container - I can make the data live locally on your machine if you want, but then it would be harder to 'restart' with fresh dummy data.
3. I can dockerize all other services if you want - but developing on dockerized services can be a hassle.

#### Extra info
Remove the container with `docker-compose down`
Clean your system of the image (about 1gig) with `docker system prune` or `docker system prune -a` if you are feeling really spicy

*Super rad bro*


rhinodavid:
Thanks dude. Let me dig into this a little more later. Just from the PR summary...

As far as containerizing the other services... please no.

I'm thinking of this scenario: I want to work on the client, or the server, or even a mobile client. I don't want to use the prod db because maybe I want to add a field to Resource or something, and I'm going to add the selector for that field to the provider portal or I'm going to add the display for that field to the client.

I think we have some fake data in JSON (or the mongo dump format) and we have a script that
- starts the docker container
- starts mongo
- runs through the fake JSON and populates the db
- maybe starts [mongo express](https://github.com/mongo-express/mongo-express) in the container as well

I don't *think* we need to persist the container storage. If the person needs to make long-lived changes to the fake data, they can mutate the JSON in their branch.

jollyjerr:
Closing this for now because of architecture change.

Issue #225, opened 2020-01-09
[web] Remove notice when temperature isn't available


rhinodavid:
When the temperature API isn't available or fails to load, the web client shows a notice that the temperature isn't available. I think we can probably just hide the feature entirely.
<img width="496" alt="Screen Shot 2020-01-08 at 6 38 14 PM" src="https://user-images.githubusercontent.com/5778036/72030424-28184480-3246-11ea-96da-b42b9728faf6.png">

PR #229, opened 2020-01-16
Updt/update temp display #225

sgmullins:
This PR closes #[insert associated issue number here]

## What does this PR do?
Removes the rendering of the ! icon when weather information is not available. Shows no information when none is available. 
First Contribution!!

### DIFF
--- a/packages/upswyng-web/src/components/Temperature.tsx
+++ b/packages/upswyng-web/src/components/Temperature.tsx
@@ -36,20 +36,22 @@ const StyledLocation = styled.span`
 
 const Temperature = () => {
   const temp: undefined | null | number = useTemperature();
-  return (
-    <StyledContainer>
-      <Typography variant="srOnly" component="h2">
-        Current Temperature
-      </Typography>
-      {temp && (
+  if (!temp) {
+    return null;
+  } else {
+    return (
+      <StyledContainer>
+        <Typography variant="srOnly" component="h2">
+          Current Temperature
+        </Typography>
         <StyledTemp>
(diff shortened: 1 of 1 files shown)


packages/upswyng-web/src/components/Temperature.tsx
```diff
@@ -36,42 +34,20 @@ const StyledLocation = styled.span`
...
-            title="temperature currently unavailable"
-          >
-            <ErrorOutlineIcon fontSize="small" tabIndex={0} />
-          </StyledTooltip>
-        )}
-        {temp && <StyledDegrees>{temp}&deg;</StyledDegrees>}
-        <Typography variant="srOnly">in</Typography>
-        <StyledLocation>Boulder, CO</StyledLocation>
-      </StyledTemp>
+      {temp && (
```
jacobvenable: I think this should be applied to the entire component rather than just the temperature. At the top of this component is a "Current Temperature" heading that is visible to only screen readers and could be confusing when the temperature fails to load.

packages/upswyng-web/src/components/Temperature.tsx
```diff
@@ -36,44 +34,24 @@ const StyledLocation = styled.span`
   padding: 0;
 `;
 
-const StyledTooltip = styled(props => (
```
rhinodavid: Was this just not being used?

rhinodavid: Oh wait this was the warning thing... unregaurd

PR #235, opened 2020-01-26
feat: update the resource ID to use a slug rather than search param

jacobvenable:

## What does this PR do?

Updates the resource ID to use a slug rather than a search param value.

So instead of this URL:
`https://upswyng.org/resource?id=123456`

You'll have this URL:
`https://upswyng.org/resource/123456`


rhinodavid:
Need to run `yarn`

jacobvenable:
> Need to run `yarn` and update the lock file

Argh I thought I had. It's made me feel even more like the gif in this PR description.

PR #245, opened 2020-01-27
✨ Add actions and UI to manage user permissions

rhinodavid:
## What does this PR do?
If you are a super admin, on the "Users" page you can set other users to be admins or super admins

Screen Shot 2020-01-26 at 8 26 18 PM


rhinodavid:
Added notice:
Screen Shot 2020-01-26 at 9 48 56 PM

PR #247, opened 2020-01-27
feat: update the resource ID to use a slug rather than search param

jacobvenable:
copy of #235 . checking to see if there was a problem with that specific PR and our travis build spec.


jacobvenable:
Issue was with the Travis caching. I had tried clearing it for the former PR, but missed clearing it for the branch as well.

PR #264, opened 2020-02-04
✨ Create async work initiation and monitoring solution

rhinodavid:
Turns `upswyngbot` into a full-fledged automated worker.

- Add dependency on Redis for a work queue
-- Already added an add-on package to Heroku
- Implement work queue in Javascript with `bullmq` package
- Add logic and route to queue up and "run" a test job
- Add Slack "slash command" to create test job
- Add Websocket server to surface status of work queue
- Add endpoint and views to interface with work queue WS server


rhinodavid:
Screen Shot 2020-02-03 at 11 27 17 PM

PR #270, opened 2020-02-08
updt: adds desired back button behavior

### LINKED ISSUE #168: Back button does not take user to home page
The back button does not take the user to the home page in the app but goes back literally to the last page viewed

jollyjerr:
This PR closes #168 

#### Adds natural behavior to the back button:

1. If a user is on a resource, they are taken back to the previous page. (On the tab that they clicked the resource on).
2. If a user is browsing tabs, they are taken to the index page regardless of how many tabs they clicked on.

### DIFF
--- a/packages/upswyng-web/src/components/BackButton.tsx
+++ b/packages/upswyng-web/src/components/BackButton.tsx
@@ -29,15 +29,17 @@ const CategoryBannerArrowBack = styled(ArrowBack)`
 ` as typeof ArrowBack;
 
 interface Props {
-  backRef?: Function;
+  backButtonAction?: (() => void) | null;
 }
 
-const BackButton = ({ backRef }: Props) => {
+const BackButton = ({ backButtonAction }: Props) => {
   const history = useHistory();
 
   return (
     <CategoryBannerLink
-      onClick={() => (backRef ? backRef() : history.push("/"))}
+      onClick={() =>
+        backButtonAction ? backButtonAction() : history.push("/")
+      }
     >
       <Typography variant="srOnly">go back to previous page</Typography>
       <CategoryBannerIcon>
--- a/packages/upswyng-web/src/components/BackButton.tsx
+++ b/packages/upswyng-web/src/components/BackButton.tsx
@@ -1,7 +1,8 @@
 import { Icon, Typography } from "@material-ui/core";
+import React, { useState } from "react";
 
 import { ArrowBack } from "@material-ui/icons";
-import React from "react";
+import { Redirect } from "react-router";
 import { createBrowserHistory } from "history";
 import { font } from "../App.styles";
 import styled from "styled-components";
@@ -29,14 +30,33 @@ const CategoryBannerArrowBack = styled(ArrowBack)`
 ` as typeof ArrowBack;
 
 const customHistory = createBrowserHistory();
+const resourceRegex: RegExp = /(resource\/)/;
 
-const BackButton = () => (
-  <CategoryBannerLink onClick={() => customHistory.goBack()}>
-    <Typography variant="srOnly">go back to pervious page</Typography>
-    <CategoryBannerIcon>
-      <CategoryBannerArrowBack />
-    </CategoryBannerIcon>
-  </CategoryBannerLink>
-);
+const isOnResource = () => {
+  return createBrowserHistory().location.pathname.match(resourceRegex);
+};
+
+const BackButton = () => {
+  const [isGoingHome, setIsGoingHome] = useState(false);
+
+  const handleClick = () => {
+    if (isOnResource()) {
+      customHistory.goBack();
+    } else {
+      setIsGoingHome(true);
+    }
+  };
+
+  return isGoingHome ? (
+    <Redirect to="/" />
+  ) : (
+    <CategoryBannerLink onClick={() => handleClick()}>
+      <Typography variant="srOnly">go back to previous page</Typography>
+      <CategoryBannerIcon>
+        <CategoryBannerArrowBack />
+      </CategoryBannerIcon>
+    </CategoryBannerLink>
+  );
+};
 
 export default BackButton;
--- a/packages/upswyng-web/src/components/BackButton.tsx
+++ b/packages/upswyng-web/src/components/BackButton.tsx
@@ -1,11 +1,10 @@
 import { Icon, Typography } from "@material-ui/core";
-import React, { useState } from "react";
 
 import { ArrowBack } from "@material-ui/icons";
-import { Redirect } from "react-router";
-import { createBrowserHistory } from "history";
+import React from "react";
 import { font } from "../App.styles";
 import styled from "styled-components";
+import { useHistory } from "react-router";
 
 const CategoryBannerLink = styled.a`
   align-items: center;
@@ -29,28 +28,17 @@ const CategoryBannerArrowBack = styled(ArrowBack)`
   }
 ` as typeof ArrowBack;
 
-const customHistory = createBrowserHistory();
-const resourceRegex: RegExp = /(resource\/)/;
+interface Props {
+  backRef?: Function;
+}
 
-const isOnResource = () => {
-  return createBrowserHistory().location.pathname.match(resourceRegex);
-};
+const BackButton = ({ backRef }: Props) => {
+  const history = useHistory();
 
-const BackButton = () => {
-  const [isGoingHome, setIsGoingHome] = useState(false);
-
-  const handleClick = () => {
-    if (isOnResource()) {
-      customHistory.goBack();
-    } else {
-      setIsGoingHome(true);
-    }
-  };
-
-  return isGoingHome ? (
-    <Redirect to="/" />
-  ) : (
-    <CategoryBannerLink onClick={() => handleClick()}>
+  return (
+    <CategoryBannerLink
+      onClick={() => (backRef ? backRef() : history.push("/"))}
+    >
       <Typography variant="srOnly">go back to previous page</Typography>
       <CategoryBannerIcon>
         <CategoryBannerArrowBack />
--- a/packages/upswyng-web/src/components/Resource.tsx
+++ b/packages/upswyng-web/src/components/Resource.tsx
@@ -89,7 +89,7 @@ export const Resource = () => {
       <PageBanner
         color={resourceColor}
         text={resource.name}
-        backRef={!!lastLocation ? history.goBack : undefined}
+        backButtonAction={lastLocation ? history.goBack : null}
       />
       <List component="div">
         {resource.address && (
--- a/packages/upswyng-web/src/components/Resource.tsx
+++ b/packages/upswyng-web/src/components/Resource.tsx
@@ -22,6 +22,7 @@ import { Theme } from "@material-ui/core/styles/createMuiTheme";
 import Typography from "@material-ui/core/Typography";
 import makeStyles from "@material-ui/core/styles/makeStyles";
 import { useHistory } from "react-router";
+import { useLastLocation } from "react-router-last-location";
 import { useParams } from "react-router-dom";
 import useResource from "./useResource";
 
@@ -60,6 +61,7 @@ export const Resource = () => {
   const resource = useResource(resourceId || "");
   const listIconClasses = useListIconStyles({});
   const history = useHistory();
+  const lastLocation = useLastLocation();
 
   if (resource === undefined) {
     return <LoadingSpinner />;
@@ -87,7 +89,7 @@ export const Resource = () => {
       <PageBanner
         color={resourceColor}
         text={resource.name}
-        backRef={history.goBack}
+        backRef={!!lastLocation ? history.goBack : undefined}
       />
       <List component="div">
         {resource.address && (
--- a/packages/upswyng-web/src/components/Resource.tsx
+++ b/packages/upswyng-web/src/components/Resource.tsx
@@ -1,5 +1,6 @@
 import { Container, colors } from "../App.styles";
 import { TCategoryDefinition, categories } from "./Categories";
+
 import BannerColorContext from "./BannerColorContext";
 import Link from "@material-ui/core/Link";
 import List from "@material-ui/core/List";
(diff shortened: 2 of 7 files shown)


packages/upswyng-web/src/components/BackButton.tsx
```diff
@@ -1,7 +1,8 @@
 import { Icon, Typography } from "@material-ui/core";
+import React, { useState } from "react";
 
 import { ArrowBack } from "@material-ui/icons";
-import React from "react";
+import { Redirect } from "react-router";
```
jacobvenable: Instead of using a rendered `<Redirect/>` React router actually has a [`useHistory` hook](https://reacttraining.com/blog/react-router-v5-1/#usehistory) that will allow you to use `history.push()` instead.

Your implementation definitely works, but I think we should add to history rather than replace the current location as `<Redirect/>` does.

packages/upswyng-web/src/components/BackButton.tsx
```diff
@@ -29,14 +30,33 @@ const CategoryBannerArrowBack = styled(ArrowBack)`
 ` as typeof ArrowBack;
 
 const customHistory = createBrowserHistory();
+const resourceRegex: RegExp = /(resource\/)/;
```
jacobvenable: Rather than having the functionality of the back button be based on the current URL, what do you think about specifically passing a path to the `<BackButton/>` component?

This prop could be something like `sourceUrl` and have a default value of "/", which accounts for directly navigating to the page where the back-button is used. Otherwise, it would be passed down from wherever its source was. So if this back button is used within the `<ResourceList/>`, its `sourceUrl` would be the home page. If the back button is used on an individual `<Resource/>` page, then its `sourceUrl` would either be the list of resources by category/subcategory or the search results page.

That way, instead of coupling this back button to a URL that could change, we'll just explicitly tell it where to go based on where the user came from.

jacobvenable:
Also where is your reaction gif?! 😝

packages/upswyng-web/src/components/BackButton.tsx
rhinodavid: hook master

packages/upswyng-web/src/components/BackButton.tsx
jollyjerr: Sounds perfect. I felt really 'dirty' parsing the url like that anyway 😂

jollyjerr:
The gifs hate me 😂 .

It may be silly, but I really dug passing the desired back behavior as a prop rather than a url string (history.goBack())? It seems to me, because router already does the work for us, it is a convenient solution?

Unfortunately, this update breaks a test that I 'fixed' by providing router context, but I am getting logs that make me think it is not actually fixed. Could one of you check it out? I'm a jest newbie and I could not debug it out for the life of me.

```txt
Warning: The tag <LoadingSpinner> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.
```
[router context in tests best practice](https://testing-library.com/docs/example-react-router)
[original broken test info](https://github.com/testing-library/react-testing-library/issues/493)

jacobvenable:
> Unfortunately, this update breaks a test that I 'fixed' by providing router context, but I am getting logs that make me think it is not actually fixed. Could one of you check it out?

Totally! Looking at the test, this may have been the result of my sloppiness anyways 😅 I'll comment on the test to explain the various updates I made, but feel free to resolve those comments at any time 👍

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
```diff
@@ -1,17 +1,24 @@
...
+import { mount } from "enzyme";
 import useResource from "../useResource";
 
 jest.mock("../Categories", () => ({
   categories: {},
 }));
-jest.mock("../../utils/searchParams");
-jest.mock("../../constants");
-jest.mock("../LoadingSpinner.tsx", () => "LoadingSpinner");
+jest.mock("../LoadingSpinner.tsx", () => () => "TEST-loading-spinner");
```
jacobvenable: While the original mock works for [shallow rending](https://airbnb.io/enzyme/docs/api/shallow.html), it wouldn't work for when [mounting](https://airbnb.io/enzyme/docs/api/ReactWrapper/mount.html).

I believe that in the former, loading spinner would be rendered as a normal HTML element during mount. The warnings you were seeing was the test complaining about a random `<LoadingSpinner/>` element that was being read as an HTML element rather than a component.

The updated mock changes the render of the loading spinner to be plain text. I also prepended the text with "TEST" to signify that this render is unique to this test and not actually rendered within the application.

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
```diff
@@ -1,17 +1,24 @@
...
+jest.mock("../PageBanner", () => () => "PageBanner");
+jest.mock("../Map", () => () => "Map");
+jest.mock("../Schedule", () => "Schedule");
+jest.mock("react-router", () => ({
+  useHistory: () => undefined,
+}));
```
jacobvenable: Rather than wrap the component within a router, I updated `useHistory` to be a noop instead. That way:

- react router never comes into play (within the test)
- whenever the component calls `useHistory`, it's still a function

If we ever wanted to test rendering based on what was returned from `useHistory`, the we would update the noop to be a `jest.fn()`, e.g.
```
jest.mock("react-router", () => ({
  useHistory: jest.fn(),
}));
```

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
```diff
@@ -25,27 +32,18 @@ jest.mock("../../App.styles", () => ({
...
-const mockedUseResource = useResource as jest.Mock;
 
-describe("<Resource/>", () => {
-  mockedUseResource.mockImplementation(() => foodResource) as typeof jest.mock;
-  const wrapper = shallow(<Resource />);
+const mockUseResource = useResource as jest.Mock;
 
-  it("renders the charityname property of the resource prop object", () => {
-    expect(wrapper.find("PageBanner").prop("text")).toBe(foodResource.name);
-  });
```
jacobvenable: I removed this test (I originally wrote 😅) because it didn't really provide much value. A better option would be to either:

- actually render out the `<PageBanner/>`, which is currently prevented by the mock of it towards the top of this file
- or opt for what I did here, and don't worry about testing `<PageBanner/>` within the test of `<Resource/>`

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
```diff
@@ -25,27 +32,18 @@ jest.mock("../../App.styles", () => ({
...
+describe("<Resource/>", () => {
+  const setup = (overrides = {}) => {
+    return mount(<Resource {...overrides} />);
+  };
```
jacobvenable: Created a quick setup function that can be used to override any props we want to provide an individual render of `<Resource/>`. If we add any tests in the future that are prop-based, we can use the setup like so:

```js
it('renders lollipops equal to the passed number of lollipops', () => {
   const numLollipops = 2;
   const wrapper = renderWith({ numLollipops });
   expect(wrapper.find('lollipop')).toHaveLength(numLollipops);
})
```

so doing `renderWith({ numLollipop: 2 })` will essentially render the resource component with a `lollipop` prop that equals `2`: `<Resource numLollipops={2} />`

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
```diff
@@ -25,27 +32,18 @@ jest.mock("../../App.styles", () => ({
...
 
   it("renders a loading spinner when a resource is not loaded", () => {
-    mockedUseResource.mockImplementation(() => undefined) as typeof jest.mock;
-    const wrapper = shallow(<Resource />);
+    mockUseResource.mockReturnValueOnce(undefined);
+    const wrapper = setup();
 
-    expect(wrapper.find("LoadingSpinner").length).toBe(1);
+    expect(wrapper.text()).toBe("TEST-loading-spinner");
```
jacobvenable: With the updated mock of the loading spinner, we're now only looking for text (checking via `.text()`) rather than an actual component, which we look for with `.find()`

packages/upswyng-web/src/components/__tests__/Resource.test.tsx
jollyjerr: Nice 😎

jollyjerr:
Sweet! The updated tests look good to me. Thank you for looking into that!

packages/upswyng-web/src/components/Resource.tsx
```diff
@@ -81,7 +84,11 @@ export const Resource = () => {
 
   return (
     <Container>
-      <PageBanner color={resourceColor} text={resource.name} />
+      <PageBanner
+        color={resourceColor}
+        text={resource.name}
+        backRef={history.goBack}
```
jacobvenable: The last issue with this implementation is that jumping directly to an individual resource page will result in the back button taking them to any previous page. For example:

1. open a new tab
2. navigate to https://deploy-preview-270--upswyng.netlify.com/resource/5def5cb7b831a048f2a3deed
3. press the back button

And it will take you back to your new tab page rather than, ideally, Upswyng's home page.

What we need to do is set this set this value to either:

A. value that is passed to the `<Resource/>` component
B. `history.goBack` only when the previous page is within Upswyng

Since `<Resource/>` is render by a route found in the main `<App/> ` passing something directly to that component won't work; however, we could use [React context](https://reactjs.org/docs/context.html) to accomplish this.

I'm not totally in love with this solution, but it's the first thing that comes to mind:

1. create a new or update some existing context that sets a the `backRef` value as `history.push('/')` by default
2. update both `<ResourceList/>` and `<SearchResults/>` to consume that context a update the `backRef` to be `history.goBack()`
3. update `<PageBanner/>` to consume that context and call `backRef` as it is before.

I now remember why I used context for setting the banner color of a resource. ([example of context provider](https://github.com/CodeForBoulder/upswyng/blob/master/packages/upswyng-web/src/components/PageBanner.tsx))

The similar problem I had was that a Resource's banner color was equal to either, in order of importance:
- the color of the category that led to the resource
- the first category color it belonged to when it was navigated to from the search results or directly

The solution I went with was that whenever a banner was given a specific color (in the case of a category/subcategory page), it would [change the color context to match it](https://github.com/CodeForBoulder/upswyng/blob/master/packages/upswyng-web/src/components/PageBanner.tsx#L37). It would then ensure that future page banners would also use that color until it was reset elsewhere, as with the [search page](https://github.com/CodeForBoulder/upswyng/blob/master/packages/upswyng-web/src/components/Search.tsx#L12).

You could create some new context for the `backRef` to accomplish something similar.

@rhinodavid could I get a sanity check?

rhinodavid: seems insane

jacobvenable: > seems insane

What would a better option be? The only options I can think of to flag whether the last location was within the app is:

- React context
- some indication in the URL (query param or optional slug?)
- local storage

rhinodavid:
I've messed around with this awhile.. so complicated!

@jollyjerr I didn't really follow the reasoning for the `backRef`. I guess I kind of get it but it seems complicated.

I noticed the problem with the back button causing navigation away from the app right off the bat, so I tried to solve that first. I though I could examine the contents of `history` and see the previous page to see if it's part of upswyng, but turns out you cant just look at the previous locations. To fix this I found [`react-router-last-location`](https://github.com/hinok/react-router-last-location), which injects the last location into `context` (this may be what @jacobvenable was getting at). Conveniently, the last location ends up being `null` if the user has just loaded upswyng.

I got a little turned around by the whole subcategory "tab" thing, but the logic seems to boil down to:
- if we're on a category, go to the home page
- if we're on a subcategory, go to the category
- otherwise go back
(correct me if that isn't right)

I added the `useLocation` hook to get the current location, then just wrote some code to do that logic. Note that all of this happens in the back button component (plus modifying App.ts to have the last location context provider).

The typechecker has been quite angry with me, but maybe you can tame it. Thanks for all the hard work.

et viola:
Screen Shot 2020-02-10 at 9 13 23 PM

jollyjerr:
### Check this out:

I am using backRef<Function> because react-router tracks where the user is when they enter a resource page and already defines a function for us to use to take them back to the page they were on.

Example click paths:
```"/" -> "/search?q=water -> "/resource/6060842"```   (history.goBack() takes the user to  "/search?q=water")
```"/" -> "food/meals" -> "/resource/6060842"```   (history.goBack() takes the user to "/food/meals")

If no action is passed down, then the backRef function will default to `history.push("/")`

No matter what additions happen in the future, we could always pass functions from the history api to control the behavior of the back button:

```"/something_new" -> "/this_part_is_not_needed_again" -> "/page_to_go_back_twice_from"``` In this situation we could easily say `backRef={history.go(-2)}` or even `backRef={history.push("/food/meals")}`

These updates would be very simple and not require conditional logic or URL parsing. 
Does that sorta make sense? If it's silly we can switch for sure!

#### This update
Because react-router-last-location returns null when a user first enters the site (awesome find!), I simply prevent `<Resource/>` from passing the backRef action down if the user does not have a last location.

**Comment by jacobvenable** [comment]
> * if we're on a category, go to the home page
> * if we're on a subcategory, go to the category
> * otherwise go back
>   (correct me if that isn't right)

Just one change:
- if we're on a subcategory, go to the home page

**Comment by rhinodavid** [comment]
> * if we're on a subcategory, go to the home page

Umm wut. So if we're on a "tab" we don't go to all, just to the main page? I guess that could make sense.

File: packages/upswyng-web/src/components/BackButton.tsx [hunk 9]
```diff
@@ -28,15 +28,23 @@ const CategoryBannerArrowBack = styled(ArrowBack)`
   }
 ` as typeof ArrowBack;
 
-const customHistory = createBrowserHistory();
+interface Props {
+  backRef?: Function;
```
rhinodavid: `backRef?: () => void` or something like that

rhinodavid: Also "ref" is a weird choice as it overloads [React refs](https://reactjs.org/docs/refs-and-the-dom.html). Maybe something like `resourceGoHome`?

packages/upswyng-web/src/components/Resource.tsx
```diff
@@ -81,7 +86,11 @@ export const Resource = () => {
 
   return (
     <Container>
-      <PageBanner color={resourceColor} text={resource.name} />
+      <PageBanner
+        color={resourceColor}
+        text={resource.name}
+        backRef={!!lastLocation ? history.goBack : undefined}
```
rhinodavid: I vary rarely (never?) intentionally set something to `undefined`.  I would pass `null` here.. you might have to change the type of the prop to `() => void | null`.. not 100% sure.

PR #274, opened 2020-02-19
:sparkles: add temperature endpoint

### LINKED ISSUE #273: create temperature endpoint
The user-facing application currently retrieves the current temperature via the [Open Weather API](https://openweathermap.org/api) directly.

We should create our own endpoint that will:
- hit the Open Weather API
- cache responses from that endpoint for 10 minutes to limit number of requests sent to that API

We then need to update the [custom hook](https://github.com/CodeForBoulder/upswyng/blob/master/packages/upswyng-web/src/components/useTemperature.tsx) in the user application to use our that new endpoint.

jollyjerr:
This PR closes #273

## What does this PR do?

Adds a temperature endpoint on our server to control how often the Open Weather API is hit and updates our `useTemperature` hook to fetch weather information from our server.


jollyjerr:
This is a work in progress - it works right now but there are some things I would love to clarify tomorrow!

packages/upswyng-server/src/routes/api/weather.ts
```diff
@@ -0,0 +1,113 @@
...
+    console.error(e);
+    res.writeHead(500, {
+      "Content-Type": "application/json",
+    });
+
+    return res.end(
+      JSON.stringify({
+        message: `Error fetching weather information: ${e.message}`,
+      })
+    );
```
rhinodavid: can shorten to:
```
return res.setStatus(500).json({message: blah blah});
```
(`json` will automatically apply the `Content-Type` and stringify for you) (I just learned all this..we can refactor some endpoints)

packages/upswyng-server/src/routes/api/weather.ts
```diff
@@ -0,0 +1,113 @@
...
+  } catch (err) {
+    console.log(err);
+  }
+};
+
+class WeatherCache {
+  millisecondsToExp: number;
+  fetchFunction: typeof getCurrentWeather;
+  cache: TWeatherCurrentResponse;
+  fetchDate: Date;
```
rhinodavid: could (technically) mark these `private`

packages/upswyng-server/src/routes/api/weather.ts
```diff
@@ -0,0 +1,113 @@
...
+  constructor(fetchFunction: () => Promise<TWeatherCurrentResponse>) {
+    this.millisecondsToExp = 600000;
```
rhinodavid: consider making this an argument w/ a default value (ie `constructor( ....  , msToExpire: number = 60000)`)

packages/upswyng-server/src/routes/api/weather.ts
```diff
@@ -0,0 +1,113 @@
+import axios from "axios";
+
+const getCurrentWeather = async (
+  lat: number | null,
```
rhinodavid: super personal technique is to not abbreviate "latitude" and "longitude" because then we get a mix of `lng` `long` `lon`. For the pass-through type from the openweather API I would not modify.

jollyjerr:
Pretty major refactor but maintains the same functionality :) Let me know if I stored the Types in the wrong place!

The Open Weather API key will have to be stored serverside now.

jacobvenable:
@jollyjerr If you can make it tonight, let's go over this and get it merged 💯

packages/upswyng-server/src/routes/api/weather.ts
jollyjerr: res.status 😄

jollyjerr:
Yo @rhinodavid - can you give it one last look?

PR #277, opened 2020-02-22
feat: add hotlines page

### LINKED ISSUE #275: create hotlines page
We need a static page that contains a list of hotlines available for clients to call in cases of crisis.

- [x] create a static page contain crisis lines from [Strapp'd](https://strappd.app/#/Crisis) (you may need to select your location to see these crisis lines. See [about page](https://github.com/CodeForBoulder/upswyng/blob/master/packages/upswyng-web/src/components/About.tsx) for example of static page
   - [x] use [selectable MUI lists](https://material-ui.com/components/lists/#selected-listitem) along with [list items using a description](https://material-ui.com/components/lists/#align-list-items)
   - [x] make each link clickable to the corresponding phone number
- [x] link the tile on the home page to this static page

jollyjerr:
This PR closes #275 

## What does this PR do?

Adds a static page with emergency hotline information.

### Question:
Do we want the descriptions or not? You can see an example of the different looks here:

hotlines


packages/upswyng-web/src/components/Hotlines.tsx
```diff
@@ -1,9 +1,101 @@
+import { Container, colors } from "../App.styles";
+import { List, ListItem, ListItemText, Typography } from "@material-ui/core";
```
jacobvenable:
```suggestion
import List from "@material-ui/core/List";
```
Use [path imports with MUI](https://material-ui.com/guides/minimizing-bundle-size/#option-1) when possible to reduce our dev bundle size.

packages/upswyng-web/src/components/Hotlines.tsx
```diff
@@ -1,9 +1,101 @@
...
+    phone: "1-800-273-8255",
+  },
+  {
+    name: "National Domestic Violence Hotline",
+    description:
+      "Operating around the clock, seven days a week, confidential and free of cost, the National Domestic Violence Hotline provides lifesaving tools and immediate support to enable victims to find safety and live lives free of abuse. Callers to The Hotline at 1-800-799-SAFE (7233) can expect highly trained, experienced advocates to offer compassionate support, crisis intervention information and referral services in over 170 languages. Our phone and chat services are available to anyone who has been affected by relationship abuse, including those who are currently in abusive relationships, those who are working to heal, friends or family of victims and survivors and anyone in the community who has questions about domestic violence.",
```
jacobvenable: Ya, these are pretty wordy. What do you think about having some of the descriptions visible and having a show/more less button to reveal the rest?

packages/upswyng-web/src/components/Hotlines.tsx
```diff
@@ -1,9 +1,101 @@
...
+
+const hotlineCards = () =>
+  hotlineList.map((hotline, i) => (
+    <ListItem button key={i} onClick={() => initiatePhoneCall(hotline.phone)}>
+      <ListItemText
+        primary={hotline.name}
+        secondary={
+          <>
+            <Typography>{hotline.phone}</Typography>
+            {hotline.description}
```
jacobvenable: If we keep descriptions, I'd say go ahead and throw these in `<Typography/>` as well 👍

packages/upswyng-web/src/components/Hotlines.tsx
```diff
@@ -1,9 +1,101 @@
...
+    </ListItem>
+  ));
+
 const Hotlines = () => (
-  <div>
-    <p>Hotlines</p>
-  </div>
+  <Container>
+    <PageBanner color={colors.pink} text={"Hotlines"} />
+    <List>{hotlineCards()}</List>
```
jacobvenable: Personal pref, but I think you could throw that `hotlines.map` directly in the render 👍

jollyjerr:
How about this?
hotlines2

packages/upswyng-web/src/components/HotlineCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+import ListItemText from "@material-ui/core/ListItemText";
+import React from "react";
+import Typography from "@material-ui/core/Typography";
+
+const initiatePhoneCall = (phoneNumber: string): void => {
+  window.open(`tel:${phoneNumber}`);
+};
+
+interface Props {
+  hotline: {
```
rhinodavid: Consider removing wrapper object

packages/upswyng-web/src/components/Hotlines.tsx
```diff
@@ -1,9 +1,105 @@
...
 const Hotlines = () => (
-  <div>
-    <p>Hotlines</p>
-  </div>
+  <Container>
+    <PageBanner color={colors.pink} text={"Hotlines"} />
+    <List>
+      {hotlineList.map(({ name, description, phone }, i) => (
+        <HotlineCard
+          key={i}
```
jacobvenable: Probably fine in this case since this is static, but [best not to use indexes as keys](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). Name would be a better choice.

jollyjerr: Oh wow, that is good to know!

packages/upswyng-web/src/components/HotlineCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+  description: string;
+  phone: string;
+}
+
+const HotlineCard = ({ name, description, phone }: Props) => {
+  const [hideText, setHideText] = React.useState(true);
+  return (
+    <ListItem button>
+      <ListItemText
+        onClick={() => initiatePhoneCall(phone)}
```
jacobvenable:
```suggestion
        component={Link}
        to={`tel${phone}`}
```
We can import `Link` from `react-router-dom` and use `ListItem`'s `component` prop to avoid needing this extra fuction.

jollyjerr: While I totally love this suggestion, It does make it impossible to click the "more" and "less" buttons if we put the functionality on the ListItem rather than the ListItemText

jacobvenable: ah, no worries then good as is 👍

PR #278, opened 2020-02-23
feat: improve search results page

jacobvenable:

## What does this PR do?

- creates new `TAlgoliaHit` type outlining what is given to and returned by Algolia and shared by web and server packages
- creates new `useSearchParam` hook to use the current value of the search query param in the URL
- updates `useSearchResults` hook to return the `[status, data]` of search results
- removes `<form/>` from `<SearchInput/>`
- adds `<SearchForm/>` component used specifically for the logic of submitting search on the home page
- adds `<SearchResults/>` component containing both `<SearchInput/>` as well as the results of the associated search
- updates styles of the search results
- creates `<LoadingSearchResults/>` for displaying skeleton while search results load
- removes global `<a>` styles in favor of using MUI's `<Link/>`
   - this was needed to properly style the search result links

## Screenshots

_example of loading search results (the rectangles animate via [MUI skeleton](https://material-ui.com/components/skeleton/#skeleton))_
Screen Shot 2020-02-22 at 7 54 48 PM

_example of loaded search results_
Screen Shot 2020-02-22 at 7 54 18 PM


packages/upswyng-web/src/App.styles.ts
```diff
@@ -89,9 +80,6 @@ export default createGlobalStyle`
     font-family: ${font.families.openSans};
     font-size: ${font.sizes.basePercent};
   }
-  a {
-    ${aStyles}
-  }
```
jacobvenable: Removing these led to needing update styles elsewhere in the app.

packages/upswyng-types/src/index.ts
```diff
@@ -1,5 +1,12 @@
 import { TTimezoneName as TTimezoneName_ } from "./TTimezoneName";
 
+export interface TAlgoliaHit {
+  objectID: string;
```
rhinodavid: consider adding a comment that for resources `objectID` = `resourceId`

jacobvenable: [updated](https://github.com/CodeForBoulder/upswyng/pull/278/commits/efb6ef5dbc5165c1ca97401eb5aee1b1d3c49358)!!

ID is ID meme

PR #300, opened 2020-03-18
enforce Yarn version

jacobvenable:

## What does this PR do?

Enforces the yarn version within our project to be 1.22.1 via [Yarn policies](https://classic.yarnpkg.com/en/docs/cli/policies/#toc-policies-set-version).

If we ever want to update the yarn version, we can do so by updating via:

```bash
yarn policies set-version [VERSION HERE]
```


jacobvenable:
I was hoping this would be straight forward but I'm having difficulty updating the Travis config to work with a policy. I'll re-open this as a draft and keep at it

PR #301, opened 2020-03-18
enforce yarn version

jacobvenable:
This PR closes #[insert associated issue number here]


jacobvenable:
I don't think we want to commit the entire yarn file to enforce a version.

PR #326, opened 2020-04-20
✨Create an event log for when an Alert goes live

rhinodavid:
Also includes documentation of the Event Log framework.


packages/upswyng-server/EVENTLOG.md
```diff
@@ -0,0 +1,154 @@
...
+
+To start, add the kind name to the `EventLogKind` object in `TEventLog.ts`:
+
+```typescript
+export const EventLogKind = {
+  // .....
+  alert_live: null,
+  // .....
+};
+export type TEventLogKind = keyof typeof EventLogKind;
```
jacobvenable, line 31: You can loop through enum values as well, right? If so, is there a benefit to this approach over using an enum?

packages/upswyng-types/src/TEventLog.ts
```diff
@@ -79,7 +87,7 @@ export type TEventLogDetail =
  */
 export interface TEventLog {
   _id: string;
-  actor: TUser;
+  actor: TUser; // the user who initiated the action which caused th eevent
```
jacobvenable, line 90:
```suggestion
  actor: TUser; // the user who initiated the action which caused the event
```

packages/upswyng-server/EVENTLOG.md
```diff
@@ -0,0 +1,154 @@
...
+There are just two simple steps to do to enable messaging to Slack.
+
+First we'll add a `case` to the `switch` in the `createTextForEventLog`
+function in `utility/slackbot.ts` to translate our Event Log to the
+message that will be sent to Slack. Check out some of the other `case`s to get an idea of what the logic should look like. Note that if you want to add an emoji you use the Slack format `:<emoji-name>:`. Check out the Slack docs [here](https://www.livexlive.com/live-events/festival/music-lives?utm_source=twitter&utm_medium=social&utm_campaign=event-musiclives2020). One other note: Slack uses [custom markup](https://api.slack.com/reference/surfaces/formatting) to add style and links to text.
+
+Next, add a call to `postEventLogMessage` from `utility/slackbot.ts` where appropriate, most commonly just after where you created and saved the Event Log.
+
+> If you don't want `upswyngbot` to send a message to Slack for your event you still need to add a `case`\
+> to the `switch` statement from above. It should throw an error because it shouldn't ever get called.
```
jacobvenable, line 154: So are you saying that if I create an event that I don't care to send to Slack, you still need to add a case that prevents the default throwing an error? If so, what should the do-nothing case look like?

packages/upswyng-server/EVENTLOG.md
```diff
@@ -0,0 +1,154 @@
...
+UpSwyng server has a framework used to record business events happening in the system
+like Resources being created or Alerts going live. Events are recorded in the database
+and are shown in the UI accessable by administrators. When an event is recorded the
+`upswyngbot` on Slack can also send a message to its group.
+
+**See [PR #326](https://github.com/CodeForBoulder/upswyng/pull/326) for the implementation of the steps in this document**
+
+## Creating a new Event Log kind
+
+For this example, we'll setup logging for when an Alert goes live. The worker job checks regularly
```
jacobvenable, line 12: Definitely separate from event docs, but scheduled workers may be a good thing to have an overview of as well.

packages/upswyng-server/EVENTLOG.md
rhinodavid: errrr lemme look at this more.. I might be pulling this from flow and enums arent exactly the same

packages/upswyng-server/EVENTLOG.md
rhinodavid: throw an error

packages/upswyng-server/EVENTLOG.md
rhinodavid: u right u right.. trying to flow it

rhinodavid:
> Some follow-ups but this documentation rocks 🚀 !! I think a link to this belongs in the server's readme as well but LGTM!

thx.. yeah no server readme yet. baby steps

PR #327, opened 2020-04-21
add provider page tabs

jacobvenable:

## What does this PR do?

- adds tabs to separate portions of the provider editor page:
   - edit provider info
   - reported issues
   - event logs
- creates reusable tab components according to [ARIA best practices](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)
- adds [Bulma Badge](https://wikiki.github.io/elements/badge/) to display number of issues for a provider
- separates the breadcrumbs of categories into separate component
- updates some use of "Resource" to "Provider

## Screenshots

tabs in action

still example of tabs

_example of a provider with no unresolved issues_
example of a provider with no unresolved issues

put it on my tab


rhinodavid:
Probs unrelated to this CL, but i have a space in between the nav bar and the hero

Screen Shot 2020-04-21 at 10 22 39 PM

packages/upswyng-server/src/routes/index.svelte
```diff
@@ -14,7 +14,6 @@
   }
 </style>
 
-`
```
rhinodavid, line 17: ummm wut

jacobvenable: Ya I'm not sure where this came from but that's where that space on the home page is coming from 😅

jacobvenable:
@rhinodavid I had actually got it working originally using a technique that you described. Admittedly, the work I did to make the tabs more reusable was partially tan excuse to learn other aspects of Svelte, but I think ease of implementation should be taken into account.

The current setup up does make adding tabs easier since you don't need to worry about conditional rendering or updating ARIA attributes. You just slap a tab and corresponding panel and you're good to go.

> i get to the bottom and there are closing tags for Tabs. I think it makes the component less flexible; as the page that's using the Tabs, ideally I would just want to know which one is selected and/or when one is pressed and then decide what to do.

The inflexibility is somewhat by design. I don't believe you're meant to have content between the tabs and tab panels; at least I don't see an example on ARIAs site where that's a thing. Maybe a middle ground would be that you could do either?

I'm not sure how to do this in Svelte yet, but if `<Tabs/>` could surface the selected tab index/view to its parent, then you could take the approach of just listening to that index instead. The panels aren't technically mandatory so they could live outside of `<Tabs/>`.

> Also the whole context/store thing is more complicated than it needs be, although I will say I havent quite figured out svelte stores to begin with.

While I was making it, I had the feeling that there was an easier way. It seemed appropriate to do both so that I had state that was specific just to the children of `<Tabs/>`, but there are prolly simplifications that could be made.

I also wanted to add some flexibility to surface some kind of `changeTab()` property, but also didn't figure out how to do that yet. Would allow for something cleaner than [this line](https://github.com/CodeForBoulder/upswyng/pull/327/files#diff-43f2b65f33cbf46584bbdd496a880df6R63).

rhinodavid:
I say go ahead and ship it then.

For the thing you talked about the `changeTab` thing.. 

So for React you would probs make a property for the component like `onChangeTab: (tabId: string) => void` or something like that.

Svelte uses an event system similar to vanilla html/js.. https://svelte.dev/tutorial/component-events

rhinodavid:
> The current setup up does make adding tabs easier since you don't need to worry about conditional rendering or updating ARIA attributes. You just slap a tab and corresponding panel and you're good to go.

Maybe one way to square all this would be to name it `TabbedContentSelector` or `TabSelectView` or so on?

jacobvenable:
> > The current setup up does make adding tabs easier since you don't need to worry about conditional rendering or updating ARIA attributes. You just slap a tab and corresponding panel and you're good to go.
> 
> Maybe one way to square all this would be to name it `TabbedContentSelector` or `TabSelectView` or so on?

I'm down. How about [TabbedContent](https://github.com/CodeForBoulder/upswyng/pull/327/commits/0ba3f120828e3acd10d3d61ac9f1f755a4a91924).

jacobvenable:
> I say go ahead and ship it then.
> 
> For the thing you talked about the `changeTab` thing..
> 
> So for React you would probs make a property for the component like `onChangeTab: (tabId: string) => void` or something like that.
> 
> Svelte uses an event system similar to vanilla html/js.. https://svelte.dev/tutorial/component-events

Ah alright so I could add `on:TabChange` as a prop to `TabbedContent` and then implement that to be dispatched on callback. I won't do that this PR since it wouldn't be used, but good to know if we take that approach later.

PR #331, opened 2020-04-25
updt: improve hotline page

### LINKED ISSUE #330: Improve hotline page accessibility
The hotline page "click to call" feature is not intuitive.

jollyjerr:
This PR closes #330

## What does this PR do?
1. Updates the way phone numbers are called so that it works on all browsers 😅 
2. Makes the phone number link intuitive 
3. Locks the "more / less" button in place so it is not floating up and down and running away from your mouse

Screen Shot 2020-04-25 at 4 53 51 PM

### DIFF
--- a/packages/upswyng-web/src/components/HotlineCard.tsx
+++ b/packages/upswyng-web/src/components/HotlineCard.tsx
@@ -5,10 +5,6 @@ import ListItemText from "@material-ui/core/ListItemText";
 import React from "react";
 import Typography from "@material-ui/core/Typography";
 
-const initiatePhoneCall = (phoneNumber: string): void => {
-  window.open(`tel:${phoneNumber.replace(/[-]/g, "")}`);
-};
-
 interface Props {
   name: string;
   description: string;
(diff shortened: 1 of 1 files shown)


packages/upswyng-web/src/components/HotlineCard.tsx
```diff
@@ -17,24 +18,31 @@ interface Props {
...
+    <ListItem>
       <ListItemText
-        onClick={() => initiatePhoneCall(phone)}
         primary={name}
         secondary={
           <>
-            <Typography>{phone}</Typography>
+            <Link
+              underline="always"
+              onClick={() => initiatePhoneCall(phone)}
```
jacobvenable, line 28:
```suggestion
              href={`tel:${phone}`}
```
Could we use the `tel:` schema here instead?

jollyjerr: Totally - we just have to include this regex so that it works on all browsers `tel:${phoneNumber.replace(/[-]/g, "")}`

jollyjerr: Totally - we just have to include this regex so that it works on all browsers tel:${phoneNumber.replace(/[-]/g, "")}

PR #345, opened 2020-05-07
♻️ [server] Standardize the build of bin scripts

rhinodavid:


rhinodavid:
I added ENV vars for slackbot API authentication for the e2e tests, and these can't be access from forked repos (which I've been using). I pushed this fork to the upstream fork (this project) and I'll pull request from there.

PR #353, opened 2020-05-09
🚚 Remove package prefix and rename "core" to "common"

rhinodavid:


package.json
```diff
@@ -79,14 +79,14 @@
...
+    "build:web": "yarn build:local-packages && yarn workspace @upswyng/web build",
     "clean": "node clean.js",
-    "heroku-postbuild": "yarn workspace @upswyng/upswyng-core build && yarn workspace @upswyng/upswyng-server build",
-    "lint": "tsc --noEmit && eslint \"packages/upswyng-*/src/**/*.{js,jsx,ts,tsx}\" --fix",
-    "lint:ci": "tsc --noEmit && eslint \"packages/upswyng-*/src/**/*.{js,jsx,ts,tsx}\" --quiet",
-    "test": "concurrently -n core,server,web -c yellow,blue,cyan \"yarn workspace @upswyng/upswyng-core test\" \"yarn workspace @upswyng/upswyng-server test\" \"yarn workspace @upswyng/upswyng-web test\"",
+    "heroku-postbuild": "yarn workspace @upswyng/common build && yarn workspace @upswyng/server build",
+    "lint": "tsc --noEmit && eslint \"packages/**/src/**/*.{js,jsx,ts,tsx}\" --fix",
+    "lint:ci": "tsc --noEmit && eslint \"packages/**/src/**/*.{js,jsx,ts,tsx}\" --quiet",
+    "test": "concurrently -n common,server,web -c yellow,blue,cyan \"yarn workspace @upswyng/common test\" \"yarn workspace @upswyng/server test\" \"yarn workspace @upswyng/web test\"",
```
jacobvenable, line 89: The `@upswyng` prefix needs removed here as well ya?

.vscode/settings.json
```diff
@@ -1,12 +1,11 @@
 {
```
jacobvenable, line 1: I hadn't realized that we had committed the vscode settings. Maybe this should be ignored.

jacobvenable:
vscode settings is an aside to this so I'll go ahead and merge so I can update my current PR

.vscode/settings.json
rhinodavid: There's literally a force include line in the gitignore.

I thought about this a bit and I think we should def include the `.vscode` stuff -- I think we should recommend VSCode in our getting started. Hopefully new people can just clone the repo, add the recommended extensions and be ready to go. It's also helpful to make sure the plugin settings match the lint rule settings.

I also have started putting launch scripts in `launch.json`. It's super easy to get the dev server up in debug mode.

PR #365, opened 2020-05-11
:sparkles: Add worker job to sync algolia index

jollyjerr:
This PR does not close an issue 🚳  

## What does this PR do?
Adds a worker job to update our Algolia index with current Resources.
Also tests out @rhinodavid 's documentation skillz.


packages/server/src/worker.ts
```diff
@@ -78,6 +85,13 @@ async function start() {
           );
         case JobKind.Test:
           return await processJobTest(job as Job<TJobTestData, TJobTestResult>);
+        case JobKind.SyncAlgolia:
+          return await processJobSyncAlgolia(
+            job as Job<TJobSyncAlgoliaData, TJobSyncAlgoliaResult>,
+            ALGOLIA_APP_ID,
```
jollyjerr, line 91: I passed these variables in to keep env variable gathering all in the same place - but I really debated this so let me know what you think is best!

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -0,0 +1,72 @@
...
+  // Initialize Algolia project and project index
+  const algolia = algoliaSearch(algoliaAppId, algoliaWriteApiKey);
+  const index = algolia.initIndex(algoliaIndexName);
+  console.info(
+    `${job.name}[${job.id}]\t: Algolia connection has been established`
+  );
+  job.updateProgress(25);
+
+  let wasUpdated = false;
+  wasUpdated = await Resource.getAll()
```
jollyjerr, line 32: This is my first time working with our database directly... so if `Resource.getAll()` was the wrong choice - let me know! Does this include the deleted resources? I could not really tell?

jollyjerr:
I referred to this documentation page for Algolia - https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/how-to/incremental-updates/?language=javascript

packages/server/src/bin/updateAlgolia.ts
```diff
@@ -1,64 +0,0 @@
-import * as dotenv from "dotenv";
```
rhinodavid, line 1: Can we keep this so we can run one-offs to sync with like the dev db

packages/server/src/worker/workerTypes.ts
```diff
@@ -57,11 +58,26 @@ export interface TJobCheckLinksResult {
...
 }
 
+/**
+ *  Update algolia index with latest resources
+ */
+export interface TJobSyncAlgoliaData {
+  kind: JobKind.SyncAlgolia;
+  userId: string; // _id of user who started this job
+}
+export interface TJobSyncAlgoliaResult {
```
rhinodavid, line 68: Is there an error from the Algolia API that we would want here?

For instance, if we try the sync to algoia and their server is down would we want to have our job succeed (presumably with `wasUpdated` `false`) and report the error back?

packages/server/src/worker/workerTypes.ts
```diff
@@ -57,11 +58,26 @@ export interface TJobCheckLinksResult {
...
+  kind: JobKind.SyncAlgolia;
+  jobName: string;
+  wasUpdated: boolean;
```
rhinodavid, line 71: would suggest a comment here in the type definition file for any params whose meaning isn't obvious.

packages/server/src/worker.ts
rhinodavid: I love it.

The only question I have is if the Algolia envs aren't there would the worker throw on start or only when a sync algolia job is pulled out of the queue. (I'm pretty sure it's on a job) The tradeoff is if it fails at start we'll def know there's something wrong. But on the other hand, we don't really want to put envs into travis, and if it fails without the envs at start then we'd have to even if we weren't testing this specifically. We don't have worker tests right now but this bit me with the server e2e tests.

Since we're not really testing the worker this is fine, but see this PR of a possible solution if interested https://github.com/CodeForBoulder/upswyng/commit/4ad4e5ad1ec971b482809e8ea1c270d0438ee4b8.

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -0,0 +1,72 @@
...
+          description,
+          subcategories: subcategories.map(s => s.name).join(","),
+        })
+      );
+      job.updateProgress(75);
+      console.info(
+        `${job.name}[${job.id}]\t: Algolia Records have been prepared. Synchronizing Algolia Index...`
+      );
+
+      return index.saveObjects(updatedAlgoliaIndex);
```
rhinodavid, line 55: What happens to algolia index items that have been deleted on our side?

jollyjerr: 😬

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -29,36 +29,59 @@ export async function processJobSyncAlgolia(
...
+            name,
+            description,
+            subcategories: subcategories.map(s => s.name).join(","),
+          };
+
+          deleted
+            ? container.delete.push(resourceIndex.objectID)
+            : container.update.push(resourceIndex);
+
+          return container;
```
rhinodavid, line 62:
```
return deleted ?
   { update: container.update, deleted: [...container.delete, resourceIndex.ObjectId] }
   :
   { update: [...container.update, resourceIndex], deleted: container.deleted };
```
```suggestion
          deleted
            ? container.delete.push(resourceIndex.objectID)
            : container.update.push(resourceIndex);

          return container;
```

rhinodavid, line 62:
```suggestion
return deleted ?
   { update: container.update, deleted: [...container.delete, resourceIndex.ObjectId] }
   :
   { update: [...container.update, resourceIndex], deleted: container.deleted };
```

rhinodavid: functionally these do the same thing, but not mutating objects is just about always a good idea.

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -29,36 +29,59 @@ export async function processJobSyncAlgolia(
   job.updateProgress(25);
 
   let wasUpdated = false;
-  Resource.getAll()
+  let errorMessage = null;
+  Resource.find({})
```
rhinodavid, line 33:
```suggestion
  Resource.getAll(/* includeDeleted = */ true)
```

rhinodavid, line 33: So for now this is great. Lets get it shipped.

Now the next level stuff:

The reason `Resource.getAll` existed but then was removed (whoops.. left the type defs.. #366 ) was because for dev it's fine, but as the dataset grows putting all the results in memory becomes unrealistic. So next level would be to break this up in batches.. fetch 10 (or 20 or 100) resources, sync them to algolia, then do the next batch. This has a few more advantages: we can provide a more realistic arg to `updateProgress` since we can call [`estimatedDocumentCount`](https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount/) on the collection at the start, then we'll know the percentage of the way through the records we are for each batch and we'll send a progress update. This also will have implications on the network side when we make the request to algolia. Smaller requests are more likely to succeed for one, and also there's probably a limit on how much data we can send to algolia at once (https://www.algolia.com/doc/guides/sending-and-managing-data/send-and-update-your-data/#batching).

On the flip side this is more complicated. And we have to worry about more failure modes. What happens if batch #4 of 27 fails when we try to push it algolia. Does our job die? Do the rest get pushed but those are missing in the index? Do we retry that failure? How quickly? How many times before we give up? Do we need to retry the whole job?

So for our tiny dataset, this is fine, but you'll def run into the need for batching sooner or later if you haven't already.

rhinodavid: Also.. this function is `async` already. I'd probably want to `await` these promises

packages/server/src/worker/workerTypes.ts
```diff
@@ -68,7 +68,8 @@ export interface TJobSyncAlgoliaData {
 export interface TJobSyncAlgoliaResult {
   kind: JobKind.SyncAlgolia;
   jobName: string;
-  wasUpdated: boolean;
+  wasUpdated: boolean; // true if the index was successfully updated; false otherwise
```
rhinodavid, line 71: I probably wouldn't use `wasUpdated` here. It takes a comment to explain and I feel like `succeeded` would be more than enough for readers to understand.

There's another issue here -- is it the case that if `wasUpdated` is `false` the `errorMessage` will. be populated, and visa versa? If that's true then we're using two fields to tell us one thing. Also, I kinda led you down the wrong path here I think. Perhaps we should just throw if `index.saveObjects` or `index.deleteObjects` fails. We'll see the failure in the worker queue for sure... but on the other hand if we catch it and save as it is now that could give us a place to retry down the road.

If it's not obvious -- API design (in this case designing our in/out types) is one of the most tricky bits to this kind of software engineering. A lot of it is an art and a lot of it you have to learn from experience (yours or someone else's). APIs are also one of the hardest things to change once they are in the wild since they are necessarily the boundary of a system and so other systems are depending on them. My team goes back and forth forever on this part; the actual implementation normally is pretty easy once we've put a lot of thought into the interface.

rhinodavid: I'd prob just throw the algolia error. Sorry your PRs have been subject to the whims of my thought process.

rhinodavid: And I'd probably put `deletedRecords: string[] // ObjectIds` and `insertedRecords: string[]` (are they inserted or updated?) in the response.

jollyjerr: They are inserted and updated depending on if Algolia has the id already :)

jollyjerr:
@rhinodavid I went ahead and migrated it to a batched job because I feel like the benefits outway the drawbacks. Algolia has a super stable setup where 3 servers would all have to be down at the same time for any requests to fail https://www.algolia.com/doc/api-client/methods/advanced/ (not saying this should never be upgraded to handle that with more grace). Could you give this one more look? Thankssssss 😄

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -0,0 +1,111 @@
...
+        toDelete: [] as string[],
+      }
+    );
+
+    console.info(
+      `${job.name}[${job.id}]\t: Algolia Records have been prepared. ${updatedAlgoliaIndex.toUpdate.length} records will be updated and ${updatedAlgoliaIndex.toDelete.length} records will be deleted. Synchronizing Algolia Index...`
+    );
+    try {
+      await Promise.all([
+        index.deleteObjects(updatedAlgoliaIndex.toDelete),
```
rhinodavid, line 86: I think the results of these will give you the no-shit IDs which were updated/deleted. _theoretically_ they should be the same as `toDelete` and `toUpdate` but it's something to consider

packages/server/src/worker/processJobSyncAlgolia.ts
```diff
@@ -0,0 +1,111 @@
...
+        index.saveObjects(updatedAlgoliaIndex.toUpdate),
+      ]);
+      succeeded = true;
+      deletedRecords = [...deletedRecords, ...updatedAlgoliaIndex.toDelete];
+      insertedRecords = [
+        ...insertedRecords,
+        ...updatedAlgoliaIndex.toUpdate.map(r => r.objectID),
+      ];
+    } catch (error) {
```
rhinodavid, line 95: This catch block does, sadly, nothing.

packages/server/src/worker/processJobSyncAlgolia.ts
jollyjerr: 👀 I like this very much

jollyjerr: I did not implement this yet because the typing for @types/algoliasearch [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/619f46c319532a2ca9d487723725e76b2f578f33/types/algoliasearch/index.d.ts#L506) does not match the modern api typing [here](https://github.com/algolia/algoliasearch-client-javascript/blob/e6dec843932bdfaac13ffa1e0a11964975499078/packages/client-search/src/methods/index/saveObjects.ts#L18). I think an upgrade to ^4 would be a good idea sometime in the future

PR #376, opened 2020-05-19
🐛 Fix routing conficts between portal and web

rhinodavid:
- move all static resources for portal under `/static` directory
- make sapper app explicitly ignore paths that don't belong to the portal
- serve react root for non-sapper routes when no other middleware applies (fixes deep linking to web client)

This is the last big cleanup piece for the routing. The user should see a 404 page from the portal when appropriate; same for the web client. This, along with #375, also seems to solve the OAuth issues (knock on wood).

### DIFF
--- a/packages/server/src/app.ts
+++ b/packages/server/src/app.ts
@@ -52,14 +52,11 @@ export default function(options: TAppOptions) {
     .use(
       compression({ threshold: 0 }),
       cors(), // TODO: Lock this down to non-admin routes
-      sirv("static", {
-        dev,
-      }),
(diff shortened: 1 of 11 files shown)


jollyjerr:
When do we want to register the [web service worker](https://github.com/CodeForBoulder/upswyng/pull/375#issue-419919777)?

rhinodavid:
> When do we want to register the [web service worker](https://github.com/CodeForBoulder/upswyng/pull/375#issue-419919777)?

It should do it on load of the React app, if we're even using it?

PR #378, opened 2020-05-20
🔊 [server] Add morgan logger middleware

rhinodavid:
Add request logger middleware when running dev server.


packages/server/src/app.ts
```diff
@@ -48,6 +49,9 @@ export default function(options: TAppOptions) {
       }
     });
   }
+
+  if (dev) app.use(morgan("dev"));
```
jacobvenable, line 53: I still don't grasp why your conditional middleware for the past logger wasn't working >.<

PR #380, opened 2020-05-21
🤖 Create upswngbot command and worker job to delete all user sessions

rhinodavid:
**Should land after #381**

When making OAuth configuration changes or if the session secret somehow got disclosed, it would be necessary to destroy all user sessions.

This PR adds a worker job to drop the session collection (Mongoose will recreate it) + an endpoint to handle the command from `upswyngbot`.

Screen Shot 2020-05-20 at 10 33 57 PM

### DIFF
--- a/packages/server/src/routes/api/bot/destroy-all-sessions.ts
+++ b/packages/server/src/routes/api/bot/destroy-all-sessions.ts
@@ -11,7 +11,7 @@ async function makeJobDestroyAllSessions(_req, res, _next, user: TUser) {
       type: "section",
       text: {
         type: "mrkdwn",
-        text: `:bomb: I'm *destorying all sessions* for you!`,
+        text: `:bomb: I'm *destroying all sessions* for you!`,
       },
     },
     {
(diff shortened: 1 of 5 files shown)


packages/server/src/worker/mq.ts
```diff
@@ -118,6 +120,20 @@ async function addJobCheckNewAlerts(
   );
 }
 
+async function addJobDestroyAllSessions(
+  name: string = getName("-"),
+  userId
+): Promise<Job<TJobDestroyAllSessionsData, TJobDestroyAllSessionsResult>> {
```
jacobvenable, line 126: These job creators are looking pretty repetitive. Would it be better to create an `addJob` utility that used generic types instead?

packages/server/src/routes/api/bot/destroy-all-sessions.ts
```diff
@@ -0,0 +1,54 @@
...
+
+async function makeJobDestroyAllSessions(_req, res, _next, user: TUser) {
+  const jobName = getName("-");
+  mq.addJobDestroyAllSessions(jobName, user._id);
+  const blocks = [
+    {
+      type: "section",
+      text: {
+        type: "mrkdwn",
+        text: `:bomb: I'm *destorying all sessions* for you!`,
```
jacobvenable, line 14:
```suggestion
        text: `:bomb: I'm *destroying all sessions* for you!`,
```

packages/server/src/worker/mq.ts
rhinodavid: what would it look like? `addJob` then you put the job name and then it makes a switch statement to customize inputs? Seems about the same. I tend to be ok with copy-paste until a beneficial abstraction pops out

PR #381, opened 2020-05-21
🔧 Redo dev server warning to accommodate new dev db

rhinodavid:


packages/server/src/server.js
```diff
@@ -31,7 +31,7 @@ const {
 
 const { ALGOLIA_INDEX_NAME } = process.env;
 
-if (dev && /heroku_.*/.test(DATABASE_NAME)) {
+if (dev && /heroku_23.*/.test(DATABASE_NAME)) {
   throw new Error(
     "💩 You're attempting to use the production datebase in a dev enviroment."
```
jacobvenable, line 36:
```suggestion
    "💩 You're attempting to use the production database in a dev enviroment."
```

packages/server/src/server.js
```diff
@@ -31,7 +31,7 @@ const {
 
 const { ALGOLIA_INDEX_NAME } = process.env;
 
-if (dev && /heroku_.*/.test(DATABASE_NAME)) {
+if (dev && /heroku_23.*/.test(DATABASE_NAME)) {
```
jacobvenable, line 34: So this is the result of the production database name changing, yes?

rhinodavid: nope its the result of the dev database changing... which makes me think nobody has actually used the dev database yet

Issue #383, opened 2020-05-21
[portal] Deleted/trashed resources showing up on subcategory page


rhinodavid:
If a non-logged in user clicks on them, they see a 404 page (only admins can see trashed resources)

PR #384, opened 2020-05-21
🐛 [server] Filter deleted resources from subcategory list

rhinodavid:
**Submit after #381**

Closes Issue #383.

so cleen

- Optionally filter out deleted resources from subcategory list

### DIFF
--- a/packages/server/src/models/Subcategory.ts
+++ b/packages/server/src/models/Subcategory.ts
@@ -125,11 +125,16 @@ SubcategorySchema.statics.getSubcategoryList = async function(
 };
 
 SubcategorySchema.statics.getByStub = async function(
-  stub: string
+  stub: string,
+  includeDeletedResources: boolean = false
 ): Promise<TSubcategoryDocument | null> {
-  return await this.findOne({ stub })
+  const result = await this.findOne({ stub })
     .populate("parentCategory")
     .populate("resources");
+  if (result && !includeDeletedResources) {
+    result.resources = result.resources.filter(r => !r.deleted);
+  }
+  return result;
 };
 
 const Subcategory = mongoose.model<TSubcategoryDocument>(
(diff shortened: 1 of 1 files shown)


packages/server/src/models/Subcategory.ts
```diff
@@ -125,11 +125,16 @@ SubcategorySchema.statics.getSubcategoryList = async function(
 };
 
 SubcategorySchema.statics.getByStub = async function(
-  stub: string
+  stub: string,
+  includeDeletedResources: boolean = false
```
jacobvenable, line 129: Is it best practice to provide a type when you have a default value?

packages/server/src/models/Subcategory.ts
```diff
@@ -125,11 +125,16 @@ SubcategorySchema.statics.getSubcategoryList = async function(
...
 ): Promise<TSubcategoryDocument | null> {
-  return await this.findOne({ stub })
+  const result = await this.findOne({ stub })
     .populate("parentCategory")
     .populate("resources");
+  if (result && !includeDeletedResources) {
+    result.resources = result.resources.filter(r => !r.deleted);
+  }
```
jacobvenable, line 136: Could we incorporate the filter in the query rather than mapping over all results?

rhinodavid: It's a good idea, but the problem is the the query is for the subcategory. The subcategory already has references to the deleted resources (and the non-deleted ones).

rhinodavid: As a side note, the whole populate thing is problematic since, depending on whether it's called, the Subcategory Document could be an array of resource IDs or an array of resource documents. I'm thinking about doing a refactor so we never call `populate` and then just do all the ID -> Type conversions at the very end.

PR #396, opened 2020-06-03
feat: conditionally render login button on provider home

### LINKED ISSUE #355: [portal] Landing page says "Log In To Add Yours" even when logged in
Screen Shot 2020-05-09 at 12 21 13 AM

jacobvenable:
This PR closes #355 

## What does this PR do?

Conditionally renders a button on the Provider portal home page:

If not logged in: 

Screen Shot 2020-06-03 at 7 39 04 AM

If logged in:

Screen Shot 2020-06-03 at 7 39 17 AM

I'm in

### DIFF
--- a/packages/server/src/routes/provider/index.svelte
+++ b/packages/server/src/routes/provider/index.svelte
@@ -1,6 +1,9 @@
 <script>
-  import { goto } from "@sapper/app";
+  import { goto, stores } from "@sapper/app";
   import ResourceSearch from "./../../components/ResourceSearch.svelte";
+
+  const { session } = stores();
+  const isLoggedIn = $session && !!$session.user;
 </script>
 
 <style>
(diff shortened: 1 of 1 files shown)


packages/server/src/routes/provider/index.svelte
```diff
@@ -1,6 +1,9 @@
 <script>
-  import { goto } from "@sapper/app";
+  import { goto, stores } from "@sapper/app";
   import ResourceSearch from "./../../components/ResourceSearch.svelte";
+
+  const { session } = stores();
+  const isLoggedIn = $session && !!$session.user;
```
jollyjerr, line 6: Does sapper handle session logic for us?

rhinodavid: yee

rhinodavid: The sapper middleware has a mechanism to pluck things out of the server-side session and make them available to rendered pages via a Svelte Store.

Doc: https://sapper.svelte.dev/docs#Seeding_session_data
Impl: https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/server.js#L125

rhinodavid: To close the loop ->

There's session middleware on the server to store sessions in mongodb (we could've used redis for this too). The Oauth/Grant middleware sets session data and handles the cookie and such, so that the request on the server gets populated with session data (currently just the logged in user). Then the sapper middleware pulls it out so you can access it on svelte pages.

rhinodavid: Actually the cookie part is the session middleware

jollyjerr: That is sweeeeeeeet stuff

PR #411, opened 2020-08-26
Add safe parking subcategory under shelter option.

LisaHJung:
## What does this PR do?
Adds a safe parking subcategory under shelter option
fantastic

### DIFF
--- a/packages/web/src/components/Categories.tsx
+++ b/packages/web/src/components/Categories.tsx
@@ -195,6 +195,10 @@ export const categories: Record<TCategoryName, TCategoryDefinition> = {
     },
     subCategories: [
       {
+        text: "Safe Parking",
+        stub: "safe-parking",
+      },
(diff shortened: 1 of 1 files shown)


packages/web/src/components/Categories.tsx
```diff
@@ -194,6 +194,10 @@ export const categories: Record<TCategoryName, TCategoryDefinition> = {
       stub: "shelters",
     },
     subCategories: [
+      {
+        text: "Safe Parking",
+        stub: "safe-parking",
+      },
```
jacobvenable, line 200: Mike got back to me about ordering. Let's put this between "Abused" and "Pregnant" sub-categories. After that's done, this is good to go!

LisaHJung: Copy that. It's done!

PR #416, opened 2020-09-07
:sparkles: add ability to search for icons when creating an alert

### LINKED ISSUE #414: add wind alert icon
Wind can be a significant issue for those without shelter. UpSwyng provider portal admins want to provide alerts about potentially hazardous conditions, and have a corresponding icon. 

Since we use [Font Awesome](https://fontawesome.com/icons/) for icons,  their [wind icon (https://fontawesome.com/icons/wind) will work perfectly.

The addition can be made [here](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/provider/alert/create.svelte#L41).

jollyjerr:
This PR closes #414 

## What does this PR do?

1. Adds an interface to query the FontAwesome GraphQL API for useable icons and select one for an alert.
ezgif com-video-to-gif

Screen Shot 2020-09-06 at 7 23 06 PM

2. Adds the wind Icon as a default option

### DIFF
--- a/packages/server/src/routes/provider/alert/create.svelte
+++ b/packages/server/src/routes/provider/alert/create.svelte
@@ -20,6 +20,7 @@
   import ResourceEditor from "./../../../components/ResourceEditor.svelte";
   import rgbHex from "rgb-hex";
   import TimePicker from "./../../../components/TimePicker.svelte";
+  import axios from "axios";
 
   const { tz } = moment;
   const { session } = stores();
@@ -153,10 +154,15 @@
   }
 
   let isCustomIconSelectorOpen = false;
+  let searchSubmitted = false;
   let iconSearchTerm = "";
   let iconSearchResults = [];
 
-  function searchForCustomIcons() {
+  function searchForCustomIcons(event) {
+    if (event) {
+      event.preventDefault();
+    }
+
     iconSearchResults = [];
     const query = `{
       search(
@@ -166,25 +172,18 @@
       ) { id, Membership { free } }
     }`;
 
-    fetch("https://api.fontawesome.com", {
-      method: "POST",
-      body: JSON.stringify({ query }),
-      headers: {
-        "Content-Type": "application/json",
-      },
-    })
-      .then(async resp => {
-        const { data } = await resp.json();
-        iconSearchResults = data.search
+    axios
+      .post("https://api.fontawesome.com", { query })
+      .then(async ({ data }) => {
+        iconSearchResults = data.data.search
           .filter(({ Membership }) => Membership.free.includes("solid"))
           .map(icon => `fas fa-${icon.id}`);
-        if (iconSearchResults.length === 0) {
-          iconSearchResults = false;
-        }
       })
       .catch(e => {
         console.error(e);
-        iconSearchResults = false;
+      })
+      .finally(_ => {
+        searchSubmitted = true;
       });
   }
 
@@ -595,7 +594,7 @@
       {#if isCustomIconSelectorOpen}
         <section class="modal-card-body">
           <div class="icon-choiced-container">
-            {#if iconSearchResults}
+            {#if iconSearchResults.length}
               <div class="icon-choices">
                 {#each iconSearchResults as icon}
                   <div class="icon-container">
@@ -604,7 +603,9 @@
                       class:has-background-info={tempIcon === icon}
                       on:click={() => {
                         tempIcon = icon;
-                        icons[icon] = icon;
+                        if (!Object.values(icons).includes(icon)) {
+                          icons[icon] = icon;
+                        }
                       }}>
                       <span class="icon is-large">
                         <i class={`${icon} fa-2x`} />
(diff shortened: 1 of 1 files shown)


packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -546,36 +587,105 @@
...
+                        <i class={`${icon} fa-2x`} />
+                      </span>
+                    </button>
+                  </div>
+                {/each}
+              </div>
+            {:else}
+              <div>
+                <h5>No results found</h5>
               </div>
```
jacobvenable, line 619: Instead of using an `h5` here, maybe we can spruce it up with [Bulma's notification component](https://bulma.io/documentation/elements/notification/). Minus the delete icon in their example, I think the `is-info` modification would work for this case.

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -151,6 +152,42 @@
...
+      },
+    })
+      .then(async resp => {
+        const { data } = await resp.json();
+        iconSearchResults = data.search
+          .filter(({ Membership }) => Membership.free.includes("solid"))
+          .map(icon => `fas fa-${icon.id}`);
+        if (iconSearchResults.length === 0) {
+          iconSearchResults = false;
+        }
```
jacobvenable, line 183: Instead of changing the type of `iconSearchResults`, let's have it be an array and do a check of both this array and the query term to determine if we have results.

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -546,36 +587,105 @@
...
-                  class:has-background-info={tempIcon === icon}
-                  on:click={() => (tempIcon = icon)}>
-                  <span class="icon is-large">
-                    <i class={`${icon} fa-2x`} />
-                  </span>
-                </button>
+      {#if isCustomIconSelectorOpen}
+        <section class="modal-card-body">
+          <div class="icon-choiced-container">
+            {#if iconSearchResults}
```
jacobvenable, line 598:
```suggestion
            {#if iconSearchTerm && iconSearchResults.length}
```
So if we keep `iconSearchResults` as an array, we can check both the term and the array's length to determine if we're supposed to have results. There may be a little extra work to do to handle an initial load where the user has not entered a search term nor do we have search results either.

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -546,36 +587,105 @@
...
+            Save changes
+          </button>
+          <button
+            class="button is-info"
+            on:click={() => {
+              isCustomIconSelectorOpen = false;
+            }}>
+            Back
+          </button>
+        </footer>
```
jacobvenable, line 651: Can we move this footer outside the conditional so that the pre-set icons and searched icons can share it to reduce duplication? We could just have a conditional on the 2nd button that handles the button text and action depending on whether they're looking at pre-set or searching icons.

jacobvenable (commented):
Whoa, I didn't expect an icon search to be implemented but this rocks! I think Mike is going to greatly appreciate having the ability to use most icons from Font Awesome.

I have a few comments I think we should resolve before merging, but this looks good overall. Very noice!

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -151,6 +152,42 @@
...
+    iconSearchResults = [];
+    const query = `{
+      search(
+        version: "5.12.1",
+        query: "${iconSearchTerm}",
+        first: 16
+      ) { id, Membership { free } }
+    }`;
+
+    fetch("https://api.fontawesome.com", {
```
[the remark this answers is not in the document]
jollyjerr: I prefer Axios as well, but I used fetch because David is using fetch on this page as well 😲

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -151,6 +152,42 @@
...
+      method: "POST",
+      body: JSON.stringify({ query }),
+      headers: {
+        "Content-Type": "application/json",
+      },
+    })
+      .then(async resp => {
+        const { data } = await resp.json();
+        iconSearchResults = data.search
+          .filter(({ Membership }) => Membership.free.includes("solid"))
```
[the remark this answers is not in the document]
jollyjerr: Me TOO! Like why have a graphQL api if I have to filter it like this hahaha 😂

packages/server/src/routes/provider/alert/create.svelte
jollyjerr: This ended up being a little silly? I'll see what you think haha

jollyjerr: The thing that makes it weird is not showing the error message until the user has actually hit enter for the first time

packages/server/src/routes/provider/alert/create.svelte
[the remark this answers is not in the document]
jacobvenable: Oh well that's fair then, I had missed that.

packages/server/src/routes/provider/alert/create.svelte
```diff
@@ -546,26 +586,71 @@
...
-                  on:click={() => (tempIcon = icon)}>
-                  <span class="icon is-large">
-                    <i class={`${icon} fa-2x`} />
-                  </span>
-                </button>
+      {#if isCustomIconSelectorOpen}
+        <section class="modal-card-body">
+          <div class="icon-choiced-container">
+            {#if iconSearchResults.length}
+              <div class="icon-choices">
```
jacobvenable, line 598: It looks like it'd be ok for the `.icon-choices` container to render whether we have results or not. Since `iconSearchResults` is now always an array, we can remove this conditional and only have one conditional around the "No results found".

packages/server/src/routes/provider/alert/create.svelte
jacobvenable: Yep I was thinking you'd have to catch whether search was submitted but this looks good!

jacobvenable (approved):
One more conditional update but this is good to go! Thank you! 🌮 🚀

PR #419, opened 2020-09-12
fix: update the category endpoint to only return non-trashed resources

jacobvenable:

## What does this PR do?

Updates the category endpoint to not return trashed resources.

you get out of here

### DIFF
--- a/packages/server/src/models/Category.ts
+++ b/packages/server/src/models/Category.ts
@@ -95,12 +95,19 @@ CategorySchema.statics.getCategoryList = async function(): Promise<
 };
 
 CategorySchema.statics.getByStub = async function(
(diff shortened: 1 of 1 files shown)


packages/server/src/models/Category.ts
```diff
@@ -95,12 +95,19 @@ CategorySchema.statics.getCategoryList = async function(): Promise<
 };
 
 CategorySchema.statics.getByStub = async function(
-  stub: string
+  stub: string,
+  includeDeletedResources: boolean = false
```
jacobvenable, line 99: I did add this parameter incase we want to add a query string that enables fetching trashed resources as well.

PR #423, opened 2020-09-12
adds cross-env package as devDependency to web project and adds cross…

Aaron-Lathrop:
…-env command to scripts

## What does this PR do?
This PR allows for easier application set-up by installing the "cross-env" package in the web project. Without this package, some Windows users may get errors like 'PORT' is not recognized as an internal or external command.

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -21,6 +21,7 @@
     "@upswyng/common": "0.1.0",
     "algoliasearch": "^3.33.0",
     "axios": "^0.19.0",
+    "cross-env": "^7.0.2",
     "debounce": "^1.2.0",
     "google-map-react": "^1.1.4",
     "html-react-parser": "^0.11.1",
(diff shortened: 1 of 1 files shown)


packages/web/package.json
```diff
@@ -21,6 +21,7 @@
     "@upswyng/common": "0.1.0",
     "algoliasearch": "^3.33.0",
     "axios": "^0.19.0",
+    "cross-env": "^7.0.2",
```
jacobvenable, line 24: Couple things on this, it looks like this lives under `dependencies` but this should be a dev dependency since we don't want this to be bundled with the app itself. You can remove this line from the `package.json` and then:

`yarn add -D cross-env`

The `-D` will add it to dev dependencies in the `package.json` and also update the `yarn.lock` that will be part of this PR.

jacobvenable (commented):
Ran this locally and it works for me! We may need to have a similar PR for the server package as well but that can be separate. Only one change requested before merging but after it's good to go!

jacobvenable:
Hey @Aaron-Lathrop 👋 Any updates on this one? If you don't have the time to complete, I could make the updates as well if you don't mind.

I definitely think this would be useful for other and would love to get this merged!

jollyjerr:
Hey @Aaron-Lathrop ! I am going to make these changes and get this merged some time this week if you are not available - but feel free to let me know if you want to finish this up yourself! Thanks!

jollyjerr:
This PR has moved to #431 - @Aaron-Lathrop your changes are still included in the new PR I just had to move because I did not have access to your fork!

Issue #420, opened 2020-09-12
setup foundation for multi-language support


jacobvenable:
We will eventually be providing multi-language support within the web project once English to Spanish translations can be procured.

In the meantime, development can implement the foundation for that support by creating the setup with English-only.

For this setup, lets use [react-i18 Next](https://react.i18next.com/), which provides some useful [setup instructions](https://react.i18next.com/latest/using-with-hooks).

The application will render as before; however it will enable us to provide other language translations in the future.

galbwe:
@jacobvenable can I grab this issue?

jacobvenable:
Aww ya!! I put it into in progress and assigned!

PR #427, opened 2020-10-03
Setup i18n-react for internationalization

galbwe:
This PR closes [issue 420](https://github.com/CodeForBoulder/upswyng/issues/420)

## What does this PR do?
- Adds i18n config file
- Provides placeholder translations for home buttons (full disclosure: I used Google Translate)

To test that the setup works as intended, open `i18n.ts` and in the object passed to `i18n.init`, set `lng` to `"es"` and `debug` to `true`. When you run the `web` development server, the home button text should be translated to Spanish.

## How does this PR make you feel? [:link:](http://giphy.com/categories/emotions/) 🤗  > 🤕 

image


packages/web/src/components/HomeButtons.tsx
```diff
@@ -11,17 +11,20 @@ import {
...
 import { HomeRouterLink } from "./HomeLink";
 import React from "react";
 import { THomeButtonRouterLink } from "../webTypes";
 import { colors } from "@upswyng/common";
+import { useTranslation } from "react-i18next";
 
 export const routerLinkButtons: THomeButtonRouterLink[] = [
   {
-    text: "Food",
+    text: "Home",
```
galbwe, line 26: Typo. I'll revert this.

galbwe:
I couldn't get the ci pipeline to run without errors after adding translations to `web/public`, so I put them in `i18n.ts` for now.

galbwe:
Tried using `i18next-chained-backen`, `i18next-localstorage-backend`, and `i18next-http-backend`. Got the same "file watcher limit exceeded" error in the ci pipeline. Reverted to hard coded translations. Other options are to load translations from common or set up an HTTP endpoint to serve the translation files to the HTTP backend https://github.com/i18next/i18next-http-backend

packages/web/src/i18n.ts
```diff
@@ -0,0 +1,51 @@
+import i18n from "i18next";
+import { initReactI18next } from "react-i18next";
+
+i18n
+  .use(initReactI18next) // passes i18n down to react-i18next
+  .init({
+    resources: {
+      en: {
```
jacobvenable, line 8: I'm good with keeping these here for now 👍 We'll look into moving these into the common package so we can circumvent the weird CI issue.

packages/web/src/index.tsx
```diff
@@ -18,7 +20,12 @@ const tagManagerArgs = {
 
 TagManager.initialize(tagManagerArgs);
 
-ReactDOM.render(<App />, document.getElementById("root"));
+ReactDOM.render(
+  <Suspense fallback={null}>
```
jacobvenable, line 24: Was this addition only needed when we were loading the translations from separate files?

packages/web/src/components/HomeButtons.tsx
```diff
@@ -104,57 +116,64 @@ export const routerLinkButtons: THomeButtonRouterLink[] = [
...
-        <HomeButton buttonColor={coordinatedEntryButton.color}>
-          {coordinatedEntryButton.text}
-          {coordinatedEntryButton.icon}
-        </HomeButton>
-      </Box>
+              <HomeButton
+                buttonColor={button.color}
+                data-test="home-router-button"
+              >
+                {t(button.translationKey, button.text)}
```
jacobvenable, line 151: So is `button.text` here acting as a backup if a translation key has not been supplied? I'm having trouble finding documentation on since the [second argument looks like it's supposed to be options](https://www.i18next.com/overview/api#t).

packages/web/src/index.tsx
galbwe: Thanks for catching that. It isn't needed now that the translations are hardcoded.

packages/web/src/components/HomeButtons.tsx
galbwe: Yeah, I meant to use it as a backup. Here's the [documentation reference](https://www.i18next.com/translation-function/essentials#passing-a-default-value). It looks like this also works, and is more intention revealing `t(button.translationKey, { defaultValue: button.text })`.

jacobvenable: Ah sweet! Ya either was good with me but just couldn't find where that was 😅

jacobvenable (approved):

jacobvenable:
Looks great :D awesome first contribution to UpSwyng!

PR #429, opened 2020-10-14
chore: update db connection config

jacobvenable:

## What does this PR do?

We're migrating our DB to a different host, and that comes with some slight changes in how we connect:

- the database name is now part of the connection URL
- the prod DB detection test needs to be updated

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -26,11 +26,11 @@ const { DATABASE_PASSWORD, DATABASE_URL, DATABASE_USER } = process.env;
 
 const { ALGOLIA_INDEX_NAME } = process.env;
 
-// if (dev && /heroku_23.*/.test(DATABASE_NAME)) {
-//   throw new Error(
-//     "💩 You're attempting to use the production database in a dev environment."
-//   );
-// }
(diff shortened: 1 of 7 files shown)


packages/server/src/server.js
```diff
@@ -80,7 +75,6 @@ const grantConfig = {
 mongoose
   .connect(DATABASE_URL, {
     useNewUrlParser: true,
-    dbName: DATABASE_NAME,
     user: DATABASE_USER,
     pass: DATABASE_PASSWORD,
   })
```
jacobvenable, line 80: It's slightly easier setup to have the dbName as part of the database URL when copying from the MongoDB Atlas.

jollyjerr (approved):
Hey @jacobvenable this all looks good to me - do you know of any side effects these changes will have in our deployment environments? I can go poke around, but I figure you already know if this is good to merge or not

jacobvenable:
> Hey @jacobvenable this all looks good to me - do you know of any side effects these changes will have in our deployment environments? I can go poke around, but I figure you already know if this is good to merge or not

Yep so all this does is remove the `DATABASE_NAME` env which is now included in that `DATABASE_URL` instead. Should be good to go now!

PR #434, opened 2020-12-27
:zap: update algolia sync

### LINKED ISSUE #433: Sync algolia index each time a draft is approved
Currently - if a new draft resource is approved someone has to manually sync our Algolia index for it to show up in search results on the provider portal or frontend. This is crucial for providers to update information because it is challenging to navigate to a resource otherwise.

We need to kick off a sync job automatically whenever a new draft resource is approved

jollyjerr:
This PR closes #433 

## What does this PR do?

This PR makes it so that our Algolia index is synced with new/edited resources when a draft is approved.

This Pr also adds the endpoint that upswyngbot can use to kick off a full sync job. (I'll add the command to upswyngbot once David can get me access)

This PR also bumps Axios because of its new security vulnerability.

### DIFF
--- a/packages/server/src/routes/api/resource/draft/approve/[_id].ts
+++ b/packages/server/src/routes/api/resource/draft/approve/[_id].ts
@@ -72,14 +72,21 @@ export async function post(req, res, next) {
       const updatedResourceObject = await resourceDocumentToResource(
         updatedResource
       );
-      await algoliaIndex.saveObject({
-        objectID: updatedResourceObject._id,
-        name: updatedResourceObject.name,
-        description: updatedResourceObject.description,
-        subcategories: updatedResourceObject.subcategories
-          .map(s => s.name)
-          .join(","),
-      });
+      console.log(updatedResourceObject.deleted);
+      if (updatedResourceObject.deleted) {
+        console.log(`deleting ${updatedResourceObject._id}`);
+        await algoliaIndex.deleteObject(updatedResourceObject._id);
+      } else {
+        console.log(`updating ${updatedResourceObject._id}`);
+        await algoliaIndex.saveObject({
+          objectID: updatedResourceObject._id,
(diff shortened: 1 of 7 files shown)


jollyjerr:
Hey @jacobvenable, I updated this so that resources are updated or added whenever a draft is approved

packages/server/src/utility/algoliaIndex.ts
```diff
@@ -0,0 +1,22 @@
+import * as dotenv from "dotenv";
+
+import algoliaSearch from "algoliasearch";
+
+dotenv.config();
+
+const {
+  ALGOLIA_APP_ID,
+  ALGOLIA_WRITE_API_KEY,
```
jacobvenable, line 9: I created a new key for this and added to Heroku. We can confirm that key works after this is deployed.

packages/server/src/routes/api/resource/draft/approve/[_id].ts
```diff
@@ -67,6 +68,24 @@ export async function post(req, res, next) {
     );
     await DraftResource.deleteByRecordId(draftToApprove._id);
 
+    try {
+      const updatedResourceObject = await resourceDocumentToResource(
+        updatedResource
+      );
+      await algoliaIndex.saveObject({
```
Comment (jacobvenable) [inline packages/server/src/routes/api/resource/draft/approve/[_id].ts:75]: We'll want an extra conditional to handle when a draft soft deletes the resource instead.

Comment (jollyjerr) [reply packages/server/src/routes/api/resource/draft/approve/[_id].ts:75]: Good point, I didn't consider that deletion also happens from this endpoint. I'll update with that asap

jollyjerr:
Hey @jacobvenable I think I broke the record for longest outstanding PR with this one haha. It should be good to go now - I tested it against the replica index we have and It was sending the appropriate requests (though obviously not actually impacting the index)

jacobvenable (approved):

PR #435, opened 2021-01-14
Favoriting resources

### LINKED ISSUE #33: implement ability for users to add favorite services
We want to enable users to add listings to a favorites list for easy future access.

## Technical Considerations

- we don't provide any type of user-login, so favorites will need to be stored via local storage
- an endpoint will need created that accepts a list of resource IDs 

### Add to Favorites

On an [individual listing](http://beta.upswyng.org/resource/5def5cb7b831a048f2a3deed), display a [floating action button](https://material-ui.com/components/floating-action-button/#floating-action-button) that adds/removes an item from the user's favorites.

- [ ] when favorited, display a [primary orange](https://github.com/CodeForBoulder/upswyng/blob/master/packages/common/src/colors.ts#L13) favorite icon `import FavoriteIcon from '@material-ui/icons/Favorite';`
- [ ] when not favorited, display a [primary orange](https://github.com/CodeForBoulder/upswyng/blob/master/packages/common/src/colors.ts#L13) favorite border icon `import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';`
- [ ] when pressing the button, it toggles whether the resource is favorited or unfavorited
- [ ] when the user has favorited an item for the first time, display an info [snackbar](https://material-ui.com/components/snackbars/#snackbar) that displays the following message for five seconds "Favorites are only saved to this device."

### Favorites Page

- [ ] create a new API endpoint that accepts a list of resource IDs and returns the info needed to display favorites in a view similar to a [category/subcategory list](http://beta.upswyng.org/food)
- [ ] create a favorites page of a similar design to a [category/subcategory list](http://beta.upswyng.org/food)
   - no filtering of resources beyond favorites is needed
   - use the [primary orange](https://github.com/CodeForBoulder/upswyng/blob/master/packages/common/src/colors.ts#L13) color for the banner which contains the text "Favorites"
   - [ ] below the banner but above the favorites list, display an [info alert](https://material-ui.com/components/alert/#alert) with the following text: "Favorites are only saved to this device. They will not be saved if you are using a public computer, or when you have a private session enabled in your browser (incognito mode)."
   - [ ] when no items are favorited, display the following text instead of the favorites list: "You haven't added any favorites yet!"

### Main Menu Link

- [ ] add a "Favorites" link to the favorites page in the main menu using the favorite icon `import FavoriteIcon from '@material-ui/icons/Favorite';`

paspheeris:
This PR closes #33 

## What does this PR do?
This PR implements the functionality described in #33. In the process of implementing that functionality, I made two (main) changes to existing code:

1. I changed the [resources endpoint](https://github.com/CodeForBoulder/upswyng/blob/2eae3b3cde5a8c62274c0dfaf875323a24f0c14a/packages/server/src/routes/api/resources/index.ts#L1) to optionally accept an `id` param so that it can be used to query for a specific subset of resources, instead of just all resources. One thing I noticed is that unlike the [endpoint for getting a single, specific resource](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/resource/%5BresourceId%5D.ts#L19), this endpoint doesn't call `resourceDocumentToResource`. I noticed this because with resources fetched from this endpoint, no map shows up in the web frontend (there are no `latitude` and `longitude` fields on `TResourceDocument`). I am now running `resourceDocumentToResource` on the results inside the `/resources` endpoint, but let me know if I've misunderstood and this endpoint really should return `TResourceDocument`s instead of `TResource`s.
2. I changed [`useResource`](https://github.com/CodeForBoulder/upswyng/blob/2eae3b3cde5a8c62274c0dfaf875323a24f0c14a/packages/web/src/components/useResource.tsx#L1) to `useResources`, and also changed the hook to use `react-query` in the same way that `useResourcesByCategory` and `useResourcesBySubcategory` do. I  felt like this was a good change because it fixed the way the request status [was being handled implicitly via `null` and `undefined`](https://github.com/CodeForBoulder/upswyng/blob/2eae3b3cde5a8c62274c0dfaf875323a24f0c14a/packages/web/src/components/Resource.tsx#L71). This does mean results are cached now in the same way that they are in the other hooks, which I think is fine (good, even), but let me know if I've overlooked some effect that could have.


packages/web/src/components/useResources.tsx
```diff
@@ -0,0 +1,35 @@
...
+  resources: TResource[] | undefined;
+  status: "loading" | "error" | "success";
+} => {
+  const queryParam = "?id=" + resourceIds.join(",");
+
+  const { data, status } = useQuery(queryParam, getResources, {
+    staleTime: 900000, // 15 min
+  });
+
+  return { resources: data, status };
```
jacobvenable, line 32: I'd be good just returning the entire result of the `useQuery` hook 👍

packages/web/src/components/ReportIssue.tsx
```diff
@@ -38,7 +38,7 @@ const useStyles = makeStyles(() =>
 
 const ReportIssue = () => {
   const { resourceId } = useParams();
-  const resource = useResource(resourceId || "");
+  const resource = useResources([resourceId || ""]).resources?.[0];
```
jacobvenable, line 41:
```suggestion
  const resource = useResources([resourceId]).resources?.[0];
```
Missed this the first go around, but this route only renders when `resourceId` is defined in the URL 😅

packages/web/src/components/MenuDrawer.tsx
```diff
@@ -81,6 +87,15 @@ const MenuDrawer = ({ handleMenuClose, open }: MenuDrawerProps) => {
               <ListItemIcon>{PolicyIcon}</ListItemIcon>
               <ListItemText>Privacy policy</ListItemText>
             </ListItem>
+            <ListItem
+              button
+              component={RouterLink}
+              onClick={() => handleMenuClose()}
+              to="/favorite-resources"
```
jacobvenable, line 94: Sorry, missed in the issue but maybe just `/favorites`?

packages/web/src/components/FavoriteResources.tsx
```diff
@@ -0,0 +1,48 @@
...
+      return <Typography>You haven&apos;t added any favorites yet!</Typography>;
+    } else {
+      return <ResourceList resources={resources} status={status} />;
+    }
+  };
+
+  return (
+    <Container>
+      <PageBanner color={colors.orangePrimary}>
+        <Typography variant="h1">{"Favorites"}</Typography>
```
jacobvenable, line 36:
```suggestion
        <Typography variant="h1">Favorites</Typography>
```

packages/web/src/components/FavoriteResourceFAB.tsx
```diff
@@ -0,0 +1,73 @@
...
+      if (!favoriteResources) {
+        setSnackbarShowing(true);
+      }
+
+      setFavoriteResources((favoriteResources || []).concat(resourceId));
+    }
+  };
+
+  return (
+    <div>
```
jacobvenable, line 52: I think a fragment would work here so we prevent an unnecessary element.

packages/web/src/components/useResources.tsx
```diff
@@ -0,0 +1,35 @@
+import { TResource } from "@upswyng/types";
+import { TResourcePayload } from "../webTypes";
+import apiClient from "../utils/apiClient";
+import { useQuery } from "react-query";
+
+const getResources = async (queryParam: string): Promise<TResource[]> => {
```
jacobvenable, line 6: Ya, these updates are wayyyyyyyyy better 🎉

packages/web/src/components/useLocalStorage.tsx
```diff
@@ -0,0 +1,18 @@
+import { Dispatch, SetStateAction, useEffect, useState } from "react";
+
+function useLocalStorage<T>(
```
jacobvenable, line 3: pretty slick

packages/server/src/routes/api/resources/index.ts
```diff
@@ -1,8 +1,17 @@
...
   try {
-    const resources = await Resource.getAll();
+    const idParam = _req.query.id;
+    const resourceIds = idParam?.split(",")?.map(ObjectId.createFromHexString);
+    const resourceDocuments = resourceIds
+      ? await Resource.getByResourceIds(resourceIds)
+      : await Resource.getAll();
+    const resources = await Promise.all(
+      resourceDocuments.map(resourceDocumentToResource)
+    );
```
jacobvenable, line 13: Yep I think you're correct on this.

packages/web/src/components/__tests__/Resource.test.tsx
```diff
@@ -36,7 +36,7 @@ jest.mock("../../App.styles", () => ({
   },
 }));
 
-const mockUseResource = useResource as jest.Mock;
+const mockUseResource = useResources as jest.Mock;
```
jacobvenable, line 39:
```suggestion
const mockUseResources = useResources as jest.Mock;
```

Looks like the return in this test on line 47 will need updated too:

```js
it("renders a loading spinner when a resource is loading", () => {
    mockUseResources.mockReturnValueOnce({ status: "loading" });
```

jacobvenable (commented):
Other than the one test failing and some nits, this looks great! Definitely appreciate the update to implement `useQuery` :D

packages/web/src/components/useResources.tsx
paspheeris: 77071bfb2f42414f696fb8a7245d1a78508587f6

packages/web/src/components/ReportIssue.tsx
paspheeris: Ahh thought it would be nullable, didn't realize `resourceId` is `any` here... wonder if there's a more typesafe way of handling params with `react-router-dom` 🤔

paspheeris: fd1e70477f611250b33a782525e2b3cc4189a1b5

packages/web/src/components/MenuDrawer.tsx
paspheeris: 5c5736fd15f220878bce6a238bc317e2305dee06

packages/web/src/components/FavoriteResources.tsx
paspheeris: d2eb13eb48e6e40231628c6bfe04eb780b1a8096

packages/web/src/components/FavoriteResourceFAB.tsx
paspheeris: 7e7058e7dcc54a850d8e0d32ce258449de49c2c5

packages/web/src/components/__tests__/Resource.test.tsx
paspheeris: c66df1b0bf9d707e20650551321a69d59884f473

paspheeris:
Thanks for the review. I think I've got all the issues sorted. Let me know if I missed anything!

jacobvenable (approved):
This looks great 😄  Super happy about the addition of caching and the updates to the endpoint!

PR #440, opened 2021-01-19
Non-admin draft resource view

### LINKED ISSUE #422: non-admin users can't view their own draft resources
While non-admin users are able to view a list of drafts they created, they aren't able to look at the individual draft.

draft-permissions-bug

paspheeris:
This PR closes #422 

## What does this PR do?
Renders a simple view of the draft resource for non-admin users, while hiding the admin-specific approve/delete draft buttons.

Feature/improvement ideas:
1. I think this screen should have an info banner at the top along the lines of: "Your draft resource has been submitted and is currently awaiting review by one of our administrators". I think that would be good on the screen where you create the draft as well. The process (create draft -> await approval by admin -> resource created) wasn't clear to me from the provider portal until I started looking at the code.
2. Render a `ResourceEditor` (same component used to create the draft) on this screen, so that a user can make corrections to their draft. Bit complicated because we get into needing to lock the draft while it's under review etc., but I think it could be high value because otherwise (from how I understand the process), correcting even a small typo in the initial draft is a time consuming process for the user (await approval, then submit the fix and await approval again?).

### DIFF
--- a/packages/server/src/routes/provider/resource/draft/[_id].svelte
+++ b/packages/server/src/routes/provider/resource/draft/[_id].svelte
@@ -2,64 +2,51 @@
   import { ResourceSchedule } from "@upswyng/common";
 
   export async function preload({ params, query }, { user }) {
+    let draftResource = null;
     if (!user.isAdmin) {
-      const draftsForUser = (await this.fetch("/api/resources/drafts/mine", {
-          credentials: "same-origin",
-        }).then(r => r.json())).draftResources || [];
-      const matchingDraft = draftsForUser.find(draft => draft._id === params._id);
-      if(!matchingDraft) {
-        this.error(401, "You don't have permission to view this page");
+      const draftsForUserResponse = await this.fetch("/api/resources/drafts/mine", {
+        credentials: "same-origin",
+      });
+      if (draftsForUserResponse.status !== 200) {
+        this.error(draftsForUserResponse.status, draftsForUserResponse.message);
       }
+      const {draftResources} = await draftsForUserResponse.json();
+      draftResource = draftResources.find(draft => draft._id === params._id);
+      if(!draftResource) {
+        this.error(401, "You don't have permission to view this page");
+      } 
+    } else {
+      const draftResourceResponse = await this.fetch(
+        `/api/resource/draft/${params._id}`
+      );
+      if (draftResourceResponse.status !== 200) {
+        this.error(draftResourceResponse.status, draftResourceResponse.message);
+      }
+      draftResource = (await draftResourceResponse.json()).draftResource;
     }
 
-    const resourceResponse = await this.fetch(
-      `/api/resource/draft/${params._id}`
+    // See if we have an existing resource corresponding to this draft
+    const existingResourceResponse = await this.fetch(
+      `/api/resource/${draftResource.resourceId}`
     );
+    const existingResource = existingResourceResponse.status === 200
+      ? (await existingResourceResponse.json()).resource
+      : null;
 
-    const resourceData = await resourceResponse.json();
+    return {
+      draftResource: withSchedule(draftResource),
+      existingResource: withSchedule(existingResource),
+      isAdmin: user.isAdmin,
+    };
+  }
 
-    if (resourceResponse.status !== 200) {
-      this.error(resourceResponse.status, resourceData.message);
-    } else {
-      // see if we have an existing resource corresponding to this draft
-      const existingResourceResponse = await this.fetch(
-        `/api/resource/${resourceData.draftResource.resourceId}`
-      );
-      if (existingResourceResponse.status === 404) {
-        return {
-          draftResource: {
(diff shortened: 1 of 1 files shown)


packages/server/src/routes/provider/resource/draft/[_id].svelte
```diff
@@ -2,56 +2,53 @@
...
+        credentials: "same-origin",
+      });
+      if (draftsForUserResponse.status !== 200) {
+        this.error(draftsForUserResponse.status, draftsForUserResponse.message);
+      }
+      const {draftResources} = await draftsForUserResponse.json();
+      draftResource = draftResources.find(draft => draft._id === params._id);
+      if(!draftResource) {
+        this.error(401, "You don't have permission to view this page");
+      }
```
Comment (jacobvenable) [inline packages/server/src/routes/provider/resource/draft/[_id].svelte:19]: Maybe we should just move this to the bottom of this function and supply a generic, "There was a problem locating this draft." message. Thoughts?

packages/server/src/routes/provider/resource/draft/[_id].svelte
```diff
@@ -2,56 +2,53 @@
   import { ResourceSchedule } from "@upswyng/common";
 
   export async function preload({ params, query }, { user }) {
-    if (!user || !user.isAdmin) {
-      this.error(401, "You must be an admin to access this page.");
+    let draftResource = null;
+    if (!user.isAdmin) {
```
Comment (jacobvenable) [inline packages/server/src/routes/provider/resource/draft/[_id].svelte:6]:
```suggestion
    if (!user?.isAdmin) {
```
If we supply a generic message, suggested below, whenever a draft can't be retrieved, then this could handle a user accessing a draft when they aren't logged in.

It's not a huge deal, but would avoid this weird error:
Screen Shot 2021-01-21 at 9 52 38 PM

jacobvenable:
Definitely agree with the suggestions. Created some placeholder Trello tickets to make sure we follow-up on them 🎉 

https://trello.com/c/n3pIXSp9/42-create-an-issue-for-informing-on-draft-resources
https://trello.com/c/O0vul4zA/43-determine-better-draft-process

packages/server/src/routes/provider/resource/draft/[_id].svelte
Comment (paspheeris) [reply packages/server/src/routes/provider/resource/draft/[_id].svelte:19]: If we don't care about checking to make sure the `draft_.id` is within the users' own drafts, lines 5-28 could be massively simplified to just
```js
    const draftResourceResponse = await this.fetch(
      `/api/resource/draft/${params._id}`
    );
    if (draftResourceResponse.status !== 200) {
      this.error(draftResourceResponse.status, "There was a problem locating this draft.");
    }
    const { draftResource } = await draftResourceResponse.json();
```
for both admins and non-admins.

If we want to stick with checking a users' drafts and making sure they're accessing one of their own drafts, then I think the specificity of the 401 is nice to communicate. (Will admit that the check is a bit YAGNI as I was thinking ahead to this page allowing users to modify their draft, so I'd be fine with ditching it for now in the service of code clarity).

packages/server/src/routes/provider/resource/draft/[_id].svelte
Comment (paspheeris) [reply packages/server/src/routes/provider/resource/draft/[_id].svelte:6]: Ahh good catch. Is there a route through the UI I missed that allows a non logged in user to access drafts? We can do better than a 500 in any case!

If we ditch the check as I mentioned below, then I think what you recommended is a good fix (well, this specific line would be deleted, but the `user.isAdmin` below in the return).

If we keep the check, then I think
```js      
if(!user) {
  this.error(401, "You don't have permission to view this page");
} 
```
at the very top would be the correct way to handle it.

paspheeris:
Thanks for the review. Central question to me is if we keep the check of the users' draft resources and restrict auth that way. I proposed some solutions that hopefully address the issues you brought up, depending on the answer to that question.

packages/server/src/routes/provider/resource/draft/[_id].svelte
Comment (jacobvenable) [reply packages/server/src/routes/provider/resource/draft/[_id].svelte:19]: > If we want to stick with checking a users' drafts and making sure they're accessing one of their own drafts, then I think the specificity of the 401 is nice to communicate.

Ok I can get on board with that 👍  I think it's alright to surface the drafts existence but explain that they can't access this.

packages/server/src/routes/provider/resource/draft/[_id].svelte
Comment (jacobvenable) [reply packages/server/src/routes/provider/resource/draft/[_id].svelte:6]: Ya I think we should continue checking the user is accessing one of their own drafts, so your proposal of:

1. check if there is a user
2. check that the user can access the draft
3. show the things

Works for me 👍

Comment (paspheeris) [reply packages/server/src/routes/provider/resource/draft/[_id].svelte:11]: 05a4101d9876a22cd7367313a41820fc88888a64
Alright, added that `!user` check to the top.
Also realized that I needed `return` calls below each `this.error` call... not a huge issue, but gives the correct error message that way.
I think we should be good to go now!

jacobvenable (approved):

PR #442, opened 2021-01-21
Google Maps location input

### LINKED ISSUE #388: [portal] Don't make service providers put in their latitude/longitude
Screen Shot 2020-05-22 at 3 18 05 PM

In the Service Provider editor (`ResourceEditor.svelte`) there is a "Location" field where the provider is to enter their latitude and longitude. This is probably not something the average person is going to be able to easily do.

Replace the latitude/longitude input with some other input -- I'm thinking maybe the Google location API input where you type in an address and it autocompletes. I'm guessing it's possible to get a latitude/longitude from the value of that. ([Places Autocomplete](https://developers.google.com/maps/documentation/javascript/places-autocomplete))

If this seems promising, it's probably also worth just replacing the entire address input with that component and then pulling all the data fields we need from there. If this value includes a Google ID for the place, we should also consider storing that.

paspheeris:
This PR closes #388

## What does this PR do?
It creates a `LocationInput` component which uses the Google Maps and Google Places APIs in order to autofill the address and latitude/longitude fields of the `ResourceEditor`.

Remarks:
-I've completely removed the lat/long fields from the form, but retained the address manual input in case users either can't find their provider via the search, or need to make corrections.
-There is more data, such as the phone number and business schedule, that we can grab from the Places API and potentially autofill. I think almost the whole form could be autofilled, and there are additional things that could be used to improve the client (like photos). I wanted to get #388 closed, but will draft a feature proposal for the ways this work could be continued.


packages/server/.env.example
```diff
@@ -7,6 +7,7 @@ DATABASE_PASSWORD=upswyng123
 DATABASE_SESSION_SECRET=
 DATABASE_URL=mongodb://localhost:27017/upswyng-dev
 DATABASE_USER=upswyng-dev-user
+GOOGLE_MAPS_API_KEY=<example google maps api key>
```
jacobvenable, line 10:
```suggestion
GOOGLE_PLACES_API_KEY=<example google places api key>
```
May as well be a bit more specific here.

packages/server/README.md
```diff
@@ -48,6 +48,18 @@ DATABASE_USER=upswyng-dev-user
...
+<details>
+  <summary>Instructions for configuring Google Maps APIs (optional)</summary>
+  
+*These steps are only necessary if you would like to use the location search feature on the provider portal.*
+     
+1. If you have not already created a project on the [Google Cloud Platform](https://cloud.google.com/) for use with OAuth, create a new project now.
+2. Click '+ ENABLE APIS AND SERVICES' on the API dashboard, search for 'Maps JavaScript API', then select and enable.
+3. Now enable the 'Places API' in the same way.
+4. Select 'Credentials' on the menu to the left, then '+ CREATE CREDENTIALS' then 'API Key'
+5. Copy and paste this newly created API key to the `GOOGLE_MAPS_API_KEY` field of the server `.env` file.
```
jacobvenable, line 60:
```suggestion
5. Copy and paste this newly created API key to the `GOOGLE_PLACES_API_KEY` field of the server `.env` file. 
```

jacobvenable:
Code LGTM and this is an awesome improvement! Before approving, what do you think about updating the layout to be something like this instead? The two address inputs was a bit confusing, and I figure we can bind the autocomplete to the first address line 1.

Screen Shot 2021-02-15 at 3 48 01 PM

To mock it out, I used [Bulma's columns](https://bulma.io/documentation/columns/responsiveness/). We could probably get away with not using horizontal inputs for the address too.

packages/server/src/components/LocationInput.svelte
```diff
@@ -0,0 +1,275 @@
...
+          {#if $resourceForm.zip.errors.includes('required')}
+            <p class="help is-danger">A ZIP code is required</p>
+          {/if}
+        </div>
+      </div>
+    </div>
+  </div>
+
+  {#if mapShowing}
+    <div class=" { placeSelected ? 'column' : 'hidden' } ">
```
jacobvenable, line 267: I'm good with always showing the map so we can avoid the awkward columns

https://user-images.githubusercontent.com/6598084/108302248-d2101300-7160-11eb-82f7-0105168d6058.mp4

jacobvenable (approved):
Other than always showing the map, this LGTM!

packages/server/src/components/LocationInput.svelte
paspheeris: Cool cool I'll change the map to be always showing.
I was playing around with hiding the map mostly because if someone doesn't use any of the Google suggested results (i.e. they enter the whole address manually), then the map simply sits there without updating, which I feel could be somewhat confusing to people? I guess a different fix to that might be to kind of reverse the process and update the map if they enter everything manually 🤔

paspheeris:
Will fix the map visibility tomorrow morning and then we should be good to go.
Oh and on the API keys I'm honestly not sure whether thats for 'places' or 'maps' or just GCP in general. Will double check that and commit the suggestions + update the places where the env vars are referenced if I was wrong on that one.

jacobvenable:
> Oh and on the API keys I'm honestly not sure whether thats for 'places' or 'maps' or just GCP in general. Will double check that and commit the suggestions + update the places where the env vars are referenced if I was wrong on that one.

We can and should implement an API key that is separate from the one currently being used in the react app. If places is all we need then I'm good with naming it places 👍

packages/server/src/components/LocationInput.svelte
jacobvenable: Oh ya I get that for sure 👍 I could see a user thinking that it can be interacted with, but we can save that for the user testing being put together for the provider portal.

paspheeris:
Alright I've got that env var name updated to something more accurate, I think things should be good now, let me know if I missed anything 👌

jacobvenable (approved):
All looks good! This is a huge improvement in the interface 😄

PR #444, opened 2021-02-18
Move translations to separate json files

### LINKED ISSUE #443: move localized text to common package
In #427, some basic localization was setup in the web package as a starting point for adding multi-language support.

There was one noted with placing the translated text into a separate `.json` file, which is the desired end goal.

For this issue, we'll seek to:

- [ ] place english translations in `packages/common/src/i18n/en.json`
- [ ] place spanish translations in `packages/common/src/i18n/es.json`
- [ ] replace [current translations](https://github.com/CodeForBoulder/upswyng/blob/master/packages/web/src/i18n.ts#L7) with those imported from the common package

The intended use of these is just for the `web` package for now. While not needed for this ticket, it would be good to start considering how this could be used by the `server` package too.

paspheeris:
This PR closes #443 

## What does this PR do?
Moves translations to JSON files in the common package.

Would love to look into updating the translations and getting this live/selectable on the frontend as a next step!

🙂

### DIFF
--- a/packages/common/src/i18n/en.json
+++ b/packages/common/src/i18n/en.json
@@ -1,0 +1,15 @@
+{
+  "home": {
+    "food": "Food",
+    "shelter": "Shelter",
+    "hygiene": "Hygiene",
+    "transit": "Transit",
+    "resources": "Resources",
+    "hotlines": "Hotlines",
+    "health": "Health",
+    "wifi": "Wifi",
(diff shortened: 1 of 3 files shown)


jacobvenable (approved):
I really didn't think this would be that straight forward, but I'm glad we aren't facing the same issues as before using the shared package 🤔  Everything is running just fine on the built version of the app so we're good to go!

PR #456, opened 2021-03-04
I18n spanish 448

### LINKED ISSUE #448: localize Categories and Sub-Categories
In #444 , the translations for the home page were moved to our shared package. We're now ready to enable localization to the rest of the web package.

## Translations

https://docs.google.com/document/d/1YzJdUW2tTtJxt5OsTQ5BNjwBUm-Sy2SYGYCUeT3SQB4/edit?usp=sharing

## Content

https://github.com/CodeForBoulder/upswyng/blob/master/packages/web/src/components/Categories.tsx

## Checklist
- [ ] translations for both English and Spanish were added to the [shared package](https://github.com/CodeForBoulder/upswyng/tree/master/packages/common/src/i18n)
- [ ] translation keys were substituted for the correct content

paspheeris:
This PR closes #448 

## What does this PR do?
Adds Spanish translations for categories and subcategories.

Remarks:
-I waffled a bit on how to modify the [array of categories/subcategories](https://github.com/paspheeris/upswyng/blob/a3b9e42bd4bcc7961c8cee7db92e3d5fde42a174/packages/web/src/components/Categories.tsx#L38-L293) for use with i18next. Ultimately I like removing the `text` field from `TResourceCategory` and `TResourceSubcategory` and replacing it with `translationKey`, as I think the explicitness should help prevent errors along the lines of `<p>{category.text}</p>`. These two types are only used in a few places in the web frontend only, so it's not a major change.

🇦🇷


packages/web/src/components/Categories.tsx
```diff
@@ -40,16 +40,16 @@ export const categories: Record<TCategoryName, TCategoryDefinition> = {
     color: "gold",
     placeholder: BananaIcon,
     mainCategory: {
-      text: "Food",
+      translationKey: "food",
```
jacobvenable, line 43: I agree 👍

jacobvenable (approved):
esto es asombroso! 🌮 🚀

PR #457, opened 2021-03-10
I18n spanish 450

### LINKED ISSUE #450: localize individual resource page
In #444 , the translations for the home page were moved to our shared package. We're now ready to enable localization to the rest of the web package.

## Translations

"Individual Resource Page" section of this document:
https://docs.google.com/document/d/1YzJdUW2tTtJxt5OsTQ5BNjwBUm-Sy2SYGYCUeT3SQB4/edit?usp=sharing

## Content

- https://github.com/CodeForBoulder/upswyng/blob/master/packages/web/src/components/Resource.tsx
- https://github.com/CodeForBoulder/upswyng/blob/master/packages/web/src/utils/schedule.ts

## Checklist
- [ ] translations for both English and Spanish were added to the [shared package](https://github.com/CodeForBoulder/upswyng/tree/master/packages/common/src/i18n)
- [ ] translation keys were substituted for the correct content

paspheeris:
This PR closes #450 

## What does this PR do?
Adds Spanish translations for the individual resource page.

Remarks:
-I added the `titleAccess` prop for the icons even though there's `srOnly` text for them, because I think that's what's recommended by the [material-ui docs](https://material-ui.com/components/icons/#semantic-svg-icons) and what I read [elsewhere](https://fontawesome.com/how-to-use/on-the-web/other-topics/accessibility#web-fonts-semantic). I really need to educate myself on a11y issues better 😞 
-For the schedule section, I was able to provide translations for resources that display a `WeeklySchedule`, but the `MonthlySchedule` is more problematic. The `rrule` library's support for i18n is [pretty basic](https://github.com/jakubroztocil/rrule/issues/13), only allowing you to provide word-for-word translations and not accounting for eg word order differences between languages. I'm not sure what a good solution would be here.

🇨🇴

### DIFF
--- a/packages/web/src/components/Map.tsx
+++ b/packages/web/src/components/Map.tsx
@@ -11,10 +11,12 @@ import GoogleMapReact from "google-map-react";
 import Grid from "@material-ui/core/Grid";
 import IconButton from "@material-ui/core/IconButton";
 import LoadingSpinner from "./LoadingSpinner";
+import LocationOnIcon from "@material-ui/icons/LocationOn";
 import Snackbar from "@material-ui/core/Snackbar";
 import { TResource } from "@upswyng/types";
 import { colors } from "@upswyng/common";
 import makeStyles from "@material-ui/core/styles/makeStyles";
+import { useTranslation } from "react-i18next";
 
 const boulderCoordinates = {
   lat: 40.0156852,
@@ -75,6 +77,7 @@ interface Props {
 
 const Map = ({ address, name, latitude, longitude }: Props) => {
   const classes = useStyles();
+  const { t } = useTranslation(["map"]);
   const [googleMap, setGoogleMap] = useState<any | null>(null);
   const [googleMaps, setGoogleMaps] = useState<any | null>(null);
   const [isFetchingGoogleMaps, setIsFetchingGoogleMaps] = useState<boolean>(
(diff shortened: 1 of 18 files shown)


packages/web/src/components/Map.tsx
```diff
@@ -287,6 +289,7 @@ const Map = ({ address, name, latitude, longitude }: Props) => {
           <IconButton
             onClick={() => handleDirectionButtonClick("TRANSIT")}
             color={travelMode === "TRANSIT" ? "primary" : "default"}
+            title={t("directionsByBus")}
```
jacobvenable, line 292: Def good call on adding these!

packages/web/src/components/Resource.tsx
```diff
@@ -127,11 +129,11 @@ export const Resource = () => {
         {resource.phone && (
           <ListItem component="div">
             <ListItemIcon classes={listIconClasses}>
-              <PhoneIcon />
+              <PhoneIcon titleAccess={t("phoneNumber")} />
             </ListItemIcon>
             <ListItemText>
               <Typography component="h2" variant="srOnly">
-                Phone
+                {t("phoneNumber")}
```
jacobvenable, line 136: I like your thinking here, but in this case I don't think we need to have the title on the icon. The `srOnly` variant of the following heading sets this as hidden text that is available to screen readers.

Adding a title to the icon as well means a SR would read "phone number" twice.

jacobvenable (commented):
After removing those titles from the icons, this is good to go!

packages/web/src/components/Resource.tsx
paspheeris: Good point! f4948d9cda9b3d0a5c638a5961ee256ef0cf7ed2

jollyjerr (approved):
Nice work!! I think this is good to go and I just approved your preview so all tests should pass soon and you are good to merge whenevs!

PR #439, opened 2021-01-19
adding cd to server

aliza614:
This PR closes #[insert associated issue number here]

PR #438, opened 2021-01-19
Readme withb4youbegin and serverreadmeintegrated

aliza614:
This PR closes #[insert associated issue number here]

PR #463, opened 2021-04-16
♻️ Local dev environment improvements

### LINKED ISSUE #441: Clean up development server logging
Our development environment has grown to be extremely noisy. Not only is this frustrating, but can be confusing and make it difficult for developers to discover issues with their setup. We should address each of these issues and either:

1. Fix the problem causing the warning
2. Mute the warning if it is unfixable or irrelevant

**Most of these should be fixable**

Some errors I am seeing:

- dotenv debug
- mongo client errors
- unhandled promise rejection errors (A lot of it the noise comes from whatever is supposed to be running at :6379)
Screen Shot 2021-01-19 at 6 08 23 PM

- Circular dependencies
- unused imports
Screen Shot 2021-01-19 at 6 08 11 PM

- unused css
Screen Shot 2021-01-19 at 6 07 55 PM

### LINKED ISSUE #460: Organize local development environment setup into single command
## Why

While our `yarn workspaces` monorepo setup is powerful 💪🏼 and cool 😎, it can be really confusing for new developers - and many have turned away from upswyng completely because they were not able to get it to run locally! 😲 😭 👎🏼 🚫 

#428 helped a lot with this by providing a local database and a user sign in, but we can do MORE

tenor-67357977

## How

All of the technology is in the place where we could make a single command that lives in the root directory and starts the entire project. At work, we do this with `yarn start:local` but it could be anything really.

This command will need to:
 - [ ] Install deps
 - [ ] Start the local database
 - [ ] Build local packages
 - [ ] build the client web app
 - [ ] start the server package in dev

As part of this ticket, we also need to:
 - [ ] Clean up all readmes so that this command is easily seen and other information is in package-specific readmes
 - [ ] Figure out how to handle .env files - one option could be to set necessary env vars as part of the script because they can be public and then nobody has to go set up files and can literally just run one command
 - [ ] Update readmes to limit debugging help in the file and point to helpful/relevant documentation so that we don't have to cover all edge cases here and devs can quickly go to the documentation if something is actually wrong (which will always be up to date ☺️  )

#### Other info

I think it would also be awesome if we tackled #441 as well so that new devs get _clear_ error messages and not just a bunch of noise.

jollyjerr:
This PR closes #460 and closes #441 (Except for the circular dependencies mentioned in #441 - I think we should make an additional ticket for those)

## What does this PR do?

This PR makes many enhancements targeted at improving the developer experience of upswyng. Highlights:

1. Our docker environment now stands up a local Redis instance, so the server will stop yelling that it cant connect to Redis. Currently, it's not really _used_ locally, but now the worker service actually stands up so that is cool.
2. Setting up the app locally can now all be done by running `yarn start:local`. This new script installs dependencies, builds local packages when needed, sets up `.env` files when needed, sets up a local database (with seeds) and redis instance when needed, and starts the server and web packages in development mode. Obviously, developers do not _have_ to use this script, but it abstracts nearly the entire setup process for new engineers that just want to see the thing build and get started.
3. Cleans up a lot of really small stuff that was making the server logs really noisy (unused CSS, depreciation warnings).
4. I cleaned up our vscode settings and moved to an actual [code workspace](https://code.visualstudio.com/docs/editor/workspaces) which I personally find really improves the experience of working with a monorepo like this. This change also makes it where vscode-jest actually works and is not yelling at you all the time ☺️ . I am happy to revert this change back to the old settings folders though - I know that a workspace can feel like a big change at first 😬 ! Let me know what you think @jacobvenable 
5. Cleaned up developer documentation for getting started. I also merged #439 and #438 and adapted them into this new setup.
6. Fixed our cypress setup so now devs can run the full ci locally (see the testing section in the readme).
7. Added a `yarn reset` command as an escape hatch that totally resets someone's local environment if things are just not working out.


jollyjerr:
Hey, @jacobvenable I am going to make one more change to this before it should be merged. Right now we can't run the full test suite locally because cypress does not know how to work without being on the CI server - so I'm going to get that working and then this should be good to go ☺️  !

jollyjerr:
@jacobvenable Okay _now_ I think this is ready to go! ☺️

rhinodavid:
Get it boys I'm so proud of uuu

jollyjerr:
> Just thought I'd add that I tested this branch and it worked locally on my ubuntu environment. I was still getting those redis errors:
> [server] [ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6379
> [server] at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
> 
> I was able to click each of the apps and load the data within.

@cjlynch278 I'm glad it booted up for you! You got those redis errors because your `.env` was not up to date with what this branch needs. I might add a little logging to warn that you need to delete your `.env` files.. or perhaps I'll think of another solution.... hmmm I'll keep you posted!

packages/common/package.json
```diff
@@ -24,7 +24,7 @@
   "name": "@upswyng/common",
   "scripts": {
     "build": "rollup --config",
-    "test": "jest"
```
jacobvenable, line 27: Not sure we'll ever have anything other than unit tests in the common package but this works ¯\\\_(ツ)_/¯

packages/server/README.md
```diff
@@ -34,6 +34,7 @@ For local development, you'll only need to create a copy of `.env.example` and n
 You can do this via your system's file explorer or via the following terminal command while in the repo:
 
 ```console
+cd packages/server
 cp .env.example .env
```
jacobvenable, line 38: So the `local:start` now creates one for us if one isn't created yet, ya? If that's the case, should we note that this isn't necessary if you've already run `local:start` or recommend using that instead?

jacobvenable (approved):
Some small nits here and there but works on my end 🎉  Had some issues with running `yarn start:local` before updating Docker, so it may be worth mentioning that as a troubleshooting step, but I'd be good to see if others run into the same issue first.

I think this will really help with new members getting up and running and sticking around 😄 Awesome stuff! 

thumbs up

packages/common/package.json
jollyjerr: The jest vs-code extension likes it when all of the test commands are the same across packages haha

packages/server/README.md
jollyjerr: For sure - I'll add a note

PR #472, opened 2021-05-08
I465 sort by open

### LINKED ISSUE #465: default sort by currently open
## Description

When retrieving resources by category or subcategory, our API endpoint returns the list in an indeterminate order. It would be more useful if the default sorting of that list would have open resources sorted first.

## Checklist

- [ ] when retrieving resources by a **category** or **sub-category**, the list of resources is sorted by those open first
- [ ] when retrieving resources by a **category** or **sub-category** and no resources are open, the list of resources is not sorted

## Tech Notes

- This sorting could be used in other lists, so it may make sense to create a central `sortByOpen` [utility](https://github.com/CodeForBoulder/upswyng/tree/master/packages/server/src/utility).
- [Category API Route](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/category/%5Bstub%5D.ts)
- [Category Model](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/models/Category.ts)
- [Subcategory API Route](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/subcategory/%5Bstub%5D.ts)
- [Sub-category Model](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/models/Subcategory.ts)

jollyjerr:
This PR closes #465 

## What does this PR do?
Sorts resources so that open ones are displayed first. I moved the sorting logic to the frontend because that is when we are 
guaranteed to be dealing with a flat list of resources rather than the nested structure of a `TCategory`. I think this implementation works great, but a backend implementation is in place on 44965f7!


packages/web/src/components/CategoryResults.tsx
```diff
@@ -44,7 +45,9 @@ const CategoryResults = ({
   const status = subcategoryStub
     ? subcategoryResourcesStatus
     : categoryResourcesStatus;
-  const resources = subcategoryStub ? subcategoryResources : categoryResources;
+  const resources = ResourceSchedule.sortByOpen(
```
jacobvenable, line 48: I'm not sure that this utility belongs as a method on the `ResourceSchedule` class since most methods . Thoughts on just having a generic `sortByOpen` utility instead?

jollyjerr: Totes!

PR #481, opened 2021-07-01
initial openapi file

### LINKED ISSUE #257: [server] Document API Endpoints
Some of the API endpoints (`packages/upswyng-server/src/routes/api/**`) have comments explaining what the endpoint does, what the request should look like, and what a successful and error response look like, but some don't.

We should select a standard way of documenting these ([maybe APIDoc?](https://apidocjs.com/)) and complete documentation. We could also add a script to generate an API manual from the docs.

amites:
This PR closes #257

## What does this PR do?
Creates OpenAPI doc & some grooming

![good]


package.json
```diff
@@ -83,7 +83,8 @@
     "build:local-packages": "yarn workspace @upswyng/types build && yarn workspace @upswyng/common build",
     "build:server": "yarn build:local-packages && yarn workspace @upswyng/server build",
     "build:web": "yarn build:local-packages && yarn workspace @upswyng/web build",
-    "build:local-db": "docker compose up --build -d && mongorestore --port 27018 --username upswyng-dev-user --password upswyng123 -d upswyng-dev ./packages/server/localdb/mongodump",
+    "build:local-db": "docker-compose up --build -d && npm run restore:db",
+    "restore:db": "mongorestore --port 27018 --username upswyng-dev-user --password upswyng123 -d upswyng-dev ./packages/server/localdb/mongodump",
```
amites, line 87: the `build:local-db` command is kinda broken since it doesn't wait to run the `restore` command
broke this out as a half step towards fixing -- proper would be to create a script that waits for mongodb to be ready to accept connections and and then run

in progres

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,197 @@
+openapi: 3.0.0
```
amites, line 1: to preview the HTML version you can open
https://editor.swagger.io/

and click `File` -> `Import URL`
paste in

`https://raw.githubusercontent.com/amites/upswyng/alvin/docs/packages/server/docs/openapi.yaml`

jacobvenable: Ya, I think that this is looking good to me 👍  I haven't done a lot of API documentation myself, but makes sense to use this standard. Since a good number of tools support this spec, I think this is a good base if we end up wanting something more robust. Awesome 🎉

package.json
jollyjerr: You are definitely right! I don't think we have ever had any issues with it requesting to restore before mongo was ready, but it's definitely a possibility and if it did it for you then we should get this fixed 🚀

packages/server/docs/openapi.yaml
jollyjerr: I like it! Cool that it generates curl examples
Screen Shot 2021-07-02 at 10 17 24 AM

README.md
```diff
@@ -20,7 +20,7 @@ Make sure you have these tools installed before you begin.
 
 1. Clone (or fork and clone) this repository and `cd` into it
 
-2. Make sure docker is running on your machine. You can check by running `docker ps` in your terminal
+2. Make sure docker is running on your machine. You can check by running `docker ps` in your terminal, also ensure you have [docker-compose](https://docs.docker.com/compose/install/) installed
```
jollyjerr, line 23: 👍🏼 docker-compose will be fully supported by docker itself soon, but let's add this until it's more of the standard.

package.json
jollyjerr: We could add the "wait for mongo to be ready and then restore" part to this script https://github.com/CodeForBoulder/upswyng/blob/master/start_local.sh that we have recently created so that new devs don't have to struggle to stand the project up?

jollyjerr (commented):
I think this approach is solid. Do we want to add all docs as part of this pr or break it up into multiple?

amites:
> I think this approach is solid. Do we want to add all docs as part of this pr or break it up into multiple?

I'll update this PR and migrate changes to the README and support for waiting for docker to be ready into their own branch to keep things clean

package.json
amites: 👍

amites:
heavily updated -- reverted everything not the docs file

docs file has complete or stubs for nearly every endpoint

packages/server/src/models/Resource.ts
```diff
@@ -4,7 +4,6 @@ import {
   TLegacyResource,
```
amites, line 4: removed unused import & fix typo

packages/server/src/routes/api/bot/test-job.ts
```diff
@@ -6,7 +6,7 @@ import mq from "../../../worker/mq";
 async function makeTestJob(req, res, _next, user: TUser) {
   const jobName = getName("-");
 
-  // Command can be like /testjob 35, where the second part is the delay to apply to the job
+  // Command can be like /test-job 35, where the second part is the delay to apply to the job
```
amites, line 9: the actual command includes the `-`

packages/server/src/routes/api/resources/drafts.ts
```diff
@@ -12,6 +13,19 @@ export async function get(req, res, _next) {
       "Content-Type": "application/json",
     });
```
amites, line 14: this matches the singleton and other draft responses

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+          example: 'dev.upswyng@gmail.com'
+        isAdmin:
+          type: boolean
+          default: false
+        isSuperAdmin:
+          type: boolean
+          default: false
+        name:
+          type: string
+          example: "Jacob Venable"
```
jacobvenable, line 1643:
```suggestion
          example: "Ursella UpSwyng"
```
I'm flattered XD

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
```
jacobvenable, line 1763:
```suggestion
          example: "Ursella UpSwyng"
```

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+          properties:
+            includeCancelled:
+              type: boolean
+              nullable: true
+              default: true
+            includeUnapproved:
+              type: boolean
+              nullable: true
+              default: false
+        - $ref: '#/components/schemas/StartTimeObj'
```
jacobvenable, line 1230: Will need and `end` property here as well.

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+              schema:
+                type: object
+                properties:
+                  count:
+                    type: integer
+                  alerts:
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/AlertData'
+        '404':
```
jacobvenable, line 84: Not sure how detailed we want to be, but this can also return a 500 with the type `message` string

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+          type: array
+          items:
+            type: object
+            properties:
+              text:
+                type: object
+                properties:
+                  text:
+                    type: string
+                    example: 'I found you in UpSwyng, but you have to be an administrator to tell me to run this command'
```
jacobvenable, line 1388: Thoughts on providing the "true" [successful return](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/bot/check-links.ts#L9)? It feels odd that we're returning 200s for unknown or  unauthorized users though.

jacobvenable: Ah, I see this is used as the generic bot response though 🤔 Maybe a generic "starting job" message instead?

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+              type: array
+              items:
+                allOf:
+                  - $ref: '#/components/schemas/CategoryData'
+                  - type: object
+                    properties:
+                      resources:
+                        type: array
+                        items:
+                          $ref: '#/components/schemas/Meta_id'
```
jacobvenable, line 1454: This would be an array of resources, instead of an array of resource ids.

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+                          $ref: '#/components/schemas/MetaDateTime'
+                        __v:
+                          type: integer
+                          default: 0
+        '401':
+          $ref: '#/components/responses/ErrorAuthentication'
+        '500':
+          $ref: '#/components/responses/ErrorUnknown'
+
+  /hotlines:
```
jacobvenable, line 310: We don't currently have hotlines in our DB, or at least we're not using this endpoint to reach them. Thoughts on removing this and reapplying when we are, or just leaving as-is and updating once we get to using it?

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  count:
+                    type: integer
+                    default: 12
+                  estimatedTotal:
+                    $ref: '#/components/schemas/MetaCount'
+                  resourceIssues:
```
jacobvenable, line 395: This should be the resource draft schema instead ya?

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+                  resource:
+                    $ref: '#/components/schemas/Resource'
+        '404':
+          $ref: '#/components/responses/ErrorNotFound'
+        '500':
+          $ref: '#/components/responses/ErrorUnknown'
+
+  /resources:
+    parameters:
+      - $ref: '#/components/parameters/idQueryOptional'
```
jacobvenable, line 459: This can actually be [one or many, comma separated IDs](https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/resources/index.ts#L7). Doesn't reflect quite like that with this parameter.

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+      in: query
+      description: _id for an entry
+      required: false
+      schema:
+        type: string
+        example: 5def5cb7b831a048f2a3deed
+    includeDeletedQueryOptional:
+      name: include-deleted
+      in: query
+      description: Whether to include deleted
```
jacobvenable, line 1822:
```suggestion
      description: Whether to include drafts for resources that are already deleted
```

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+                    type: array
+                    items:
+                      $ref: '#/components/schemas/ResourceDraft'
+        '500':
+          $ref: '#/components/responses/ErrorUnknown'
+
+  /resources/drafts/mine:
+    get:
+      description: 'Get an array of resources.'
+      summary: 'Get resources.'
```
jacobvenable, line 576:
```suggestion
      description: 'Get an array of drafts I created.'
```

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+        - requireLoggedIn: []
+      responses:
+        '200':
+          description: Resource Drafts
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  uncategorizedResources:
```
jacobvenable, line 590: This would be `draftResources` that are an array of drafts.
https://github.com/CodeForBoulder/upswyng/blob/master/packages/server/src/routes/api/resources/drafts/mine.ts#L38

packages/server/docs/openapi.yaml
```diff
@@ -0,0 +1,1933 @@
...
+  - name: bot
+    description: 'Endpoints for slack bot to interact with.'
+  - name: category
+    description: 'Categories'
+  - name: sub-category
+    description: 'SubCategories'
+  - name: eventlogs
+    description: 'Event logs'
+  - name: hotlines
+    description: 'Hotlines'
```
jacobvenable, line 1929:
```suggestion
  - name: hotlines
    description: 'Hotlines'
  - name: issues
    description: 'Issues reported on resources'
```
I could see issues being their own tag and tagging all the `resources/issue`(s) with it.

jacobvenable (commented):
A few minor issues, but overall looks really good! For maintaining these, I wonder there's a tool that can create schemas based off of TS types. That could make managing these a bit simpler. Awesome job!

packages/server/docs/openapi.yaml
jollyjerr: I vote we remove it until we have a need for a hotlines endpoint 👍🏼

jollyjerr (commented):
Wow, this looks like a **lot** of work! I think this will be in a good place once we can address Jacob's feedback. As for generating schemas via type definitions - a quick search did not find a good tool, but one may exist. We also could just define the return schema by referencing the type when appropriate? Like `Code: 200, Description: TResource[] & { totalFound: 800 }` etc.... This would add an extra step for an engineer to need to go reference the type - but I almost prefer that over having two sources of truth?

packages/server/docs/openapi.yaml
amites: remove just the doc and leave the endpoint?

If that's the case

Maybe add a note in the summary / description / tag that it's not currently used

Alternatively could remove the whole endpoint from the api?

packages/server/docs/openapi.yaml
amites: updated to reference the support for multiple id's in the description

packages/server/docs/openapi.yaml
amites: for now I commented out the `/hotlines` endpoint

amites:
updated based on comments from @jacobvenable thank you for the detailed review

in terms of keeping them up to date -- sadly there isn't much out of the box available as @jollyjerr found

there's a lot that could be done with a custom script -- if that's a priority I'll tackle that next vs the default sorting

packages/server/docs/openapi.yaml
jacobvenable: > for now I commented out the /hotlines endpoint

Yep, I like that plan!

jacobvenable:
> updated based on comments from @jacobvenable thank you for the detailed review
> 
> in terms of keeping them up to date -- sadly there isn't much out of the box available as @jollyjerr found
> 
> there's a lot that could be done with a custom script -- if that's a priority I'll tackle that next vs the default sorting

Yep that works for me 👍  One tool I had briefly looked into was [tsoa](https://github.com/lukeautry/tsoa). Seems nifty but is an extra layer of maintenance for our endpoints.

jacobvenable (approved):
Changes LGTM. Thanks for tackling this 🌮 🚀

PR #484, opened 2021-07-06
grooming code

amites:
This PR closes n/a

## What does this PR do?
Grooming code structure while creating docs

![good]

### DIFF
--- a/packages/server/src/routes/api/resource/issues/_setResolved.ts
+++ b/packages/server/src/routes/api/resource/issues/_setResolved.ts
@@ -56,47 +56,47 @@ export async function setResolved(resolved: boolean, req, res, next) {
         message: `Service Provider issue with _id ${_id} could not be found`,
       })
     );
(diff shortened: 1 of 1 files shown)


jollyjerr (changes requested):
Screen Shot 2021-07-16 at 7 47 34 PM

Looks like an import got messed up somewhere. This is also a **lot** to test - what if we break apart the "small" stuff (typos and stuff) from the bigger stuff like adding awaits and deleting larger chunks of code. We want to make sure that everything behaves exactly as it is - and we don't really have a lot of resources to do a full regression test right now 😅. I love what is going on in here but I think taking it in smaller steps might be best? Jacob may disagree though and that's totally fair - I'm good with anything?

amites:
removed all the logic changes adding `await` commands and left all the pure "grooming" 

added a bunch of comments to help simplify review -- let me know if ya need to break this down further

PR #491, opened 2021-09-07
fix issue 487: HomeButton heights

### LINKED ISSUE #487: fix home page tile sizes
On the [UpSwyng home page], each tile in a row is meant to stretch to fit the entire height of the row. We've noticed in some mobile screens, the "Job Training" tile isn't stretching to fill the entire row.

We'd like to update the styles of these tiles to ensure this stretching occurs.

## Example of Problem

example-tile-not-stretching

## The Goal

example-tile-stretching

dannyirwin:
This PR closes #487

## What does this PR do?
Fixes the issue where the HomeButton background would not match the height of its row if its neighbor had a greater height. 

:partying_face:

### DIFF
--- a/packages/web/src/components/HomeButton.tsx
+++ b/packages/web/src/components/HomeButton.tsx
@@ -26,6 +26,7 @@ const useStyles = makeStyles({
     textDecoration: "none",
     textTransform: "none",
     width: "100%",
+    height: "100%",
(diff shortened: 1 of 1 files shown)


jacobvenable (approved):
LGTM and thank you for your first contribution!

PR #492, opened 2021-09-23
Implement Language Select UI #490

### LINKED ISSUE #490: Implement Language Select UI
## Description

Now that we have implemented translations for the content we control, we want to enable our users to change the language manually.

## Tasks

- [ ] Create a language dropdown in the header (see screenshot below)
   - [ ] move the logo directly to the right of the main menu button
   - [ ] place language selector to the right of the temperature
- [ ] Ensure the language dropdown only has two languages available, "English" and "Español"
   - these options do not need to be localized
- [ ] When the user selects a language, all content in the application is translated to that selected language
   - This does not apply to the provider portal (anything at https://upswyng.org/provider)
   - If there is some content that isn't translated, we'll create a separate issue to fix those

example of language selector

## Tech Notes

- use the standard version of the [MUI native select](https://material-ui.com/api/native-select/)
   - For this issue, don't worry about the discrepancies between the screenshot and what MUI provides. Just focus on the styling of the header with the newly added select
- [tutorial for toggling languages](https://react.i18next.com/legacy-v9/step-by-step-guide#d-let-the-user-toggle-the-language)

Sunset05:
This PR closes #490

## What does this PR do?
This PR allows people using the website to change the language of the interface from English to Spanish


Sunset05:
I was unable to configure the openWeather Api to test how this component looks with the weather data that is currently not being rendered (I had trouble configuring the API key).
Based on Material structure I made the assumption that the weather would be directly to the left of the language selection drop down menu as shown on issue #490

jacobvenable:
Apologies, I didn't get a chance for a final look before being away for a couple weeks! I am seeing that it's stacking when the weather is present rather than being side-by-side, but otherwise this is looking good.

Screen Shot 2021-10-11 at 5 41 57 PM
 
Unfortunately, I see that we may have missed some needed translations as well, which is separate from this work. We'll need to discuss if we should hold off on merging this until those are addressed.

Sunset05:
Hey Jacob, I just added a few more commits. I brought my translation branch up to date with the current state of the CodeForBoulder Repo, fixed the problem with the stacking weather and translation components, and double checked that the functionality is still working. I think this one is ready for a code review!

Let me know if there is anything else that I need to add to this PR!

jacobvenable (approved):
Awesome to get this released and looks great! Thanks for adding this! 🌮  🚀

PR #507, opened 2021-12-16
please add banner style updates V2

AlphaLaryBah:
Version two. Thank you.

### DIFF
--- a/packages/web/src/components/PageBanner.tsx
+++ b/packages/web/src/components/PageBanner.tsx
@@ -1,10 +1,10 @@
 import BackButton from "./BackButton";
 import BannerColorContext from "./BannerColorContext";
 import Box from "@material-ui/core/Box";
-import Grid from "@material-ui/core/Grid";
 import List from "@material-ui/core/List";
 import ListItem from "@material-ui/core/ListItem";
 import ListItemIcon from "@material-ui/core/ListItemIcon";
+import ListItemText from "@material-ui/core/ListItemText";
 import React from "react";
 import { colors } from "@upswyng/common";
 
@@ -25,12 +25,12 @@ const PageBanner = ({ children, color, backButtonAction }: Props) => {
 
   return (
     <Box bgcolor={currentBannerColor || colors.black} mb={4} py={2}>
-      <List>
-        <ListItemIcon>
-          <BackButton backButtonAction={backButtonAction} />
-        </ListItemIcon>
+      <List disablePadding>
         <ListItem>
-          <Grid item>{children}</Grid>
+          <ListItemIcon>
+            <BackButton backButtonAction={backButtonAction} />
+          </ListItemIcon>
+          <ListItemText>{children}</ListItemText>
         </ListItem>
       </List>
     </Box>
--- a/packages/web/src/components/PageBanner.tsx
+++ b/packages/web/src/components/PageBanner.tsx
@@ -2,9 +2,10 @@ import BackButton from "./BackButton";
 import BannerColorContext from "./BannerColorContext";
 import Box from "@material-ui/core/Box";
 import Grid from "@material-ui/core/Grid";
+import List from "@material-ui/core/List";
+import ListItem from "@material-ui/core/ListItem";
 import React from "react";
 import { colors } from "@upswyng/common";
-
 interface Props {
   children: React.ReactNode;
   color?: string;
@@ -26,7 +27,12 @@ const PageBanner = ({ children, color, backButtonAction }: Props) => {
         <Grid item>
           <BackButton backButtonAction={backButtonAction} />
         </Grid>
-        <Grid item>{children}</Grid>
+
+        <List>
+          <ListItem>
+            <Grid item>{children}</Grid>
+          </ListItem>
+        </List>
       </Grid>
     </Box>
   );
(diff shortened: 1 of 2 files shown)


jacobvenable:
Hey, thanks for contributing! This is a change from your last PR, correct?
https://github.com/CodeForBoulder/upswyng/pull/506

If so, go ahead and close that one with a comment pointing to this update. Thanks!

packages/web/src/components/PageBanner.tsx
```diff
@@ -26,7 +27,12 @@ const PageBanner = ({ children, color, backButtonAction }: Props) => {
         <Grid item>
           <BackButton backButtonAction={backButtonAction} />
         </Grid>
-        <Grid item>{children}</Grid>
+
+        <List>
+          <ListItem>
+            <Grid item>{children}</Grid>
+          </ListItem>
+        </List>
```
jacobvenable, line 35: Ya, I think using a MUI list here is appropriate 👍 We could follow [this example](https://github.com/jakubroztocil/rrule/issues/13) and also implement the `<BackButton/>` wrapped in a `<ListItemIcon/>`. I think that would mean that we could get rid of the `<Grid/>` wrapping this component too!

AlphaLaryBah: I tried that, on this {packages/web/src/components/PageBanner.tsx} wrapping the <BackButton/> in the List and <ListItemIcon/>  but the result was not the spacing that the Issue was asking. Do you still want me to update it?

jacobvenable: Not sure 🤔 Is that due to the extra spacing that is directly on the `<BackButton/>` component?

jacobvenable: Looks like this has changed, so I'll resolve this and start a comment on the updates!

packages/web/src/components/PageBanner.tsx
```diff
@@ -2,6 +2,9 @@ import BackButton from "./BackButton";
 import BannerColorContext from "./BannerColorContext";
 import Box from "@material-ui/core/Box";
 import Grid from "@material-ui/core/Grid";
+import List from "@material-ui/core/List";
+import ListItem from "@material-ui/core/ListItem";
+import ListItemIcon from "@material-ui/core/ListItemIcon";
```
jacobvenable, line 7:
```suggestion
import List from "@material-ui/core/List";
import ListItem from "@material-ui/core/ListItem";
import ListItemIcon from "@material-ui/core/ListItemIcon";
import ListItemText from "@material-ui/core/ListItemText";
```
I think we're going to need the `ListItemText` component as well to match the [MUI List Examples](https://v4.mui.com/components/lists/).

packages/web/src/components/PageBanner.tsx
```diff
@@ -22,12 +25,14 @@ const PageBanner = ({ children, color, backButtonAction }: Props) => {
...
+        <ListItemIcon>
           <BackButton backButtonAction={backButtonAction} />
-        </Grid>
-        <Grid item>{children}</Grid>
-      </Grid>
+        </ListItemIcon>
+        <ListItem>
+          <Grid item>{children}</Grid>
+        </ListItem>
+      </List>
```
jacobvenable, line 35: Few requested updates here:

1. The `ListItemIcon` should be a child of the `<ListItem/>` based on those MUI examples
2. Since we're removing the `<Grid container/>` from this component we don't need the `<Grid item>` here either. We could replace that with `<ListItemText/>` instead. This will get us pretty close the layout we want. After removing this last `<Grid/>` I think we're good to remove that import from the top of this component as well.
3. Finally, we'll probably want to apply the [disablePadding prop](https://v4.mui.com/api/list/#props) to the `<List/>` so we have spacing close to what was there before.

jacobvenable (commented):
Hey Alpha 👋  Thanks for the updates! I'm seeing that the icon and the text are now stacked, but we do want those to remain side-by-side.

example of stacked banner

I've added some comments throughout that I think will achieve the layout we're looking for. One final item I couldn't comment on is adding an `edge="start"` prop to the BackButton's `<IconButton/>`. This will have the left side of the icon button properly align with the rest of the content on the page.

Thanks again!

packages/web/src/components/PageBanner.tsx
AlphaLaryBah: 👋  hey,
I sent you a slack message. let me know what you think. 🤔

AlphaLaryBah: hi Jacob,
I just made a new commit with update based on the new notes. let me know how it looks on your end.
thank you. 😊

jacobvenable: Awesome, looks good so far 🎉 ! I'm going resolve this and comment on anything new (if needed) separately.

packages/web/src/components/PageBanner.tsx
```diff
@@ -22,12 +25,14 @@ const PageBanner = ({ children, color, backButtonAction }: Props) => {
 
   return (
     <Box bgcolor={currentBannerColor || colors.black} mb={4} py={2}>
```
jacobvenable, line 27:
```suggestion
    <Box bgcolor={currentBannerColor || colors.black} mb={4}>
```
Looks like we'll want to reduce this as well. Apologies, I know this is nitpicky 😓

AlphaLaryBah: no worries, please no apologies needed. I will remove this `py={2}` as suggested and recommit.
😊 thank you.

jacobvenable (approved):
LGTM! Thanks for all the updates 🎉

